PR #37, opened 2022-03-25
doc: Small typo

phcreery:
since `ce.parse('\\sqrt{5} + 7^3').numericValue?.latex` was already shown having a value, I believe this is what was meant


arnog:
Yes, thank you.

PR #53, opened 2022-09-18
Fix associativity note

sritchie:
This PR fixes the description of the associativity law in the docs. Cheers!


arnog:
Yeah, that's clearer. Thanks!

PR #59, opened 2022-11-07
Fix logarithm LaTeX strings

sritchie:
I think that the old TeX examples were incorrect for these operations. Let me know if this is right or wrong, but I hope this helps!


arnog:
Yes, that's right. Good catch!

PR #83, opened 2023-06-07
first pass at adding sourceOffsets

scytacki:


arnog:
Hi @scytacki I noticed that you worked on this PR and closed it. It seems like it would be a useful contribution to merge with the mainline. Is there a reason why this should not get merged to the main branch?

Issue #73, opened 2023-01-23
"Or" Logical Operator issue?


Frizzled:
The "And" operator works as expected:
```
    let expr = ce.box(["And", ["Equal", "True", "False"], ["Equal", "True", "True"]])
    console.log(expr.evaluate().valueOf()) // false

    expr = ce.box(["And", ["Equal", "True", "True"], ["Equal", "True", "True"]])
    console.log(expr.evaluate().valueOf()) // true
```

However, the "Or" operator does not:
```
    let expr = ce.box(["Or", ["Equal", "True", "False"], ["Equal", "True", "True"]])
    console.log(expr.evaluate().valueOf()) // true

    expr = ce.box(["Or", ["Equal", "True", "True"], ["Equal", "True", "False"]])
    console.log(expr.evaluate().valueOf()) // true

    expr = ce.box(["Or", ["Equal", "True", "False"], ["Equal", "True", "False"]])
    console.log(expr.evaluate().valueOf()) // true

    expr = ce.box(["Or", ["Equal", "True", "True"], ["Equal", "True", "True"]])
    console.log(expr.evaluate().valueOf()) // true
```

svitrol:
I was browsing definitions and found this:
![obrazek](https://github.com/cortex-js/compute-engine/assets/57999718/ea6cf5a4-374b-4204-9173-30090b51fd06)

https://github.com/cortex-js/compute-engine/blob/main/src/compute-engine/latex-syntax/dictionary/definitions-logic.ts
![obrazek](https://github.com/cortex-js/compute-engine/assets/57999718/23b26d1a-bfab-4fca-a0ed-a11bf984d734)

probably not resolving situation above, but still wrong and it is operator or issue.

arnog:
@svitrol please feel free to submit a PR. I'd be happy to review and merge it.

svitrol:
@arnog sorry for not fixing this earlier :D but was focused on logical operators and didn't notice
![obrazek](https://github.com/cortex-js/compute-engine/assets/57999718/68addc3f-84dc-4ada-89cd-b49eb90ffb75)
latexTrigger: '\\operator{False}', -> prases as True
    latexTrigger: '\\mathsf{F}', -> parses as True

arnog:
Good catch. Fixed.

PR #119, opened 2023-09-21
fix: latex-syntax logic definition fixed \\operatorname{or} 'And' to …

svitrol:
https://github.com/cortex-js/compute-engine/issues/73#issuecomment-1729769722

Here is the PR. Changed one word. Hope it helps.


arnog:
Great catch! Thanks for sending the PR. I actually didn't see the problem when looking at the screenshot, but you're right, it's certainly an improvement!

Issue #123, opened 2023-10-15
NetxJS Webpack bundling fails in 0.17.0


volesen:
I have proposed a fix in #124.

## Description

NextJS `13.5.4` fails to bundle `@cortex-js/compute-engine`.
It tries to bundle a `cjs` file as an `esm` module, resulting in a missing constructor for `ComputeEngine`.

### Steps to Reproduce

A minimal example to reproduce the issues can be found on [StackBlitz](https://stackblitz.com/edit/stackblitz-starters-yorrak?file=app%2Fpage.tsx). 

### Actual Behavior

The following error message occurs in the associated minimal example.
```
Attempted import error: 'ComputeEngine' is not exported from '@cortex-js/compute-engine' (imported as 'ComputeEngine').

Import trace for requested module:
./src/app/exercises/calculus/[seed]/page.tsx
 ⨯ src/app/exercises/calculus/[seed]/page.tsx (33:15) @ eval
 ⨯ TypeError: _cortex_js_compute_engine__WEBPACK_IMPORTED_MODULE_5__.ComputeEngine is not a constructor
    at eval (./src/app/exercises/calculus/[seed]/page.tsx:43:20)
    at Exercise (./src/app/exercises/calculus/[seed]/page.tsx:42:70)
  31 | 
  32 |   const polynomial = useMemo(() => {
> 33 |     const ce = new ComputeEngine();
     |               ^
  34 | 
  35 |     return ce.box([
  36 |       "Add",
```

### Expected Behavior

I would expect Webpack to only exclusively import the cjs or esm files. 

### Environment

**MathLive version**  0.17.0

**Operating System** macOS 

**Browser**  Safari

**NextJS** 13.5.4

volesen:
Fixed in v0.18.0

PR #122, opened 2023-10-08
Support es module

Seancheey:

PR #124, opened 2023-10-15
NetxJS Webpack bundling fails in 0.17.0

volesen:
This is a possible fix for #123.

The field `"type": "module"` defines all .js files in that package scope as ES modules, which breaks webpack bundling, if commonjs files are imported. This was introduced in `0.17.0` with #122.

The approach taken in this PR to add support for a hybrid cjs and esm package is through `exports` in `packages.json`, see https://webpack.js.org/guides/package-exports/#conditions, and should work with vite through rollup.

This has been tested on NextJS `13.5.4` and Webpack 5 and bundles correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,14 @@
   "files": [
     "/dist"
   ],
+  "exports": {
+    ".": {
+        "types": "./dist/types/compute-engine.d.ts",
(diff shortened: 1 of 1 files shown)


arnog:
OK, let's give that a try. Thanks for the PR!

PR #151, opened 2024-03-11
Evaluate special values of inverse trig functions.

dcbradley:
For your consideration: here's a patch that evaluates all the same special values for inverse trig functions as are already supported for trig functions.


arnog:
Excellent idea! The PR looks good. Thanks for submitting it!

PR #152, opened 2024-03-11
When evaluating special inverse trig values, use isEqual rather than isSame

dcbradley:
Apparently, my fingers typed isSame() when I was thinking isEqual().

By the way: the comparison is redundant, since the argument to the inverse trig function is compared numerically to the key value and only if that approximately matches does it compare using isEqual(), which may just boil down to the same numerical test.  However, I thought it might be best to call isEqual() to express the formal intent to do an exact comparison here.  If the philosophy is to allow actual floating point input to match exact values in cases like this, then I think the isEqual() test should be removed.


arnog:
Ah, I see. I think the test is redundant, comparing with the floating point value is sufficient.

arnog:
OK, I've removed the test.

PR #192, opened 2024-08-29
fix: add more trig functions to compiled functions

### LINKED ISSUE #190: `compile()` fails if expression contains 'Sec'
## Description

`compile()` fails to compile a `BoxedExpression` to a js function if the expression contains a `Sec` operand

### Steps to Reproduce

```
ce.box(JSON.parse('["Sec", "x"]'))
```

### Actual Result

Throws: `Error: Unknown function Sec`

### Expected Result

Should return a js function equivalent to
```
({x: number}) => return 1 / Math.cos(x)
```

### Environment

This is not a regression

**Compute Engine version**  "^0.24.1"

rasmus-rasmus:
Closes #190


arnog:
Looks good. Thanks for your contribution!

PR #195, opened 2024-09-08
load complex.js as an es module (complex.esm.js)

holgerengels:


arnog:
Thanks for the PR! While it seems to work when building the library, I'm having some difficulties getting Jest to work. I get this error message:

```
    /Users/arno/dev/compute-engine/node_modules/complex.esm.js/dist/src/complex.js:225
    export class Complex {
    ^^^^^^

    SyntaxError: Unexpected token 'export'

    > 1 | import { Complex } from 'complex.esm.js';
        | ^
```

Any ideas?

PR #197, opened 2024-09-25
By default log should be base10

Seancheey:
Fixed a small compiling issue in cortexjs


arnog:
Good catch. Thanks for the PR!

PR #212, opened 2024-10-24
fix: preservation of (verbatim) latex

samueltlg:
Noticed that `verbatimLatex` not being saved on BoxedExpression instances created from `ce.parse` for some time. Looks as if this traceable to commits:

> cd557e5 - 'arch: changed JSON canonical format, APIs for serialization to LaTeX and MathJSON' (2024-06-25)

^Wherein there is retraction of passing of 'metadata' to `ce.box` from within `ce.parse`
And:

> e1f9e19 - 'arch' (2024-09-25)

^In which 'metadata' is retracted both as an argument to 'box' (the function) and its inner forwarding to BoxedExpression constructor calls when handling 'MathJSON object literals'.

As a guess, it appears to me that removing `metadata` as a parameter from `box` was apt due to these properties being foreseen as present on MathJSON object literals: if present at all. Looks however that in these changes that the step of checking for this meta elsewhere (i.e. object literals) is absent.
This '*fix*' is only an educated guess & seems to address the issue,

*But* this change/request also appears to **break** a single test-case within `test/compute-engine/latex-syntax/numbers.test.ts` - which I have not investigated - and thus is not suitable for merging.
(Note at the time of this request, the HEAD of main ([#115272d](https://github.com/cortex-js/compute-engine/commit/115272dd130ae69c13815d37814de57d3269abdc), there is already extant a single test failure (`test/compute-engine/functions.test.ts` -> `Apply -> Function and Hold`), which is not related to this change).

Single commit also includes an unused utility `isExpressionObject`, which was initially used, but could now be discarded.

### DIFF
--- a/src/compute-engine/boxed-expression/box.ts
+++ b/src/compute-engine/boxed-expression/box.ts
@@ -8,7 +8,11 @@ import {
   CanonicalOptions,
 } from './public';
 
-import { Expression, MathJsonIdentifier } from '../../math-json/types';
+import {
+  Expression,
+  ExpressionObject,
+  MathJsonIdentifier,
+} from '../../math-json/types';
 import { machineValue, missingIfEmpty } from '../../math-json/utils';
 import {
   isValidIdentifier,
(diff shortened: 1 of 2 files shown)


arnog:
On first reading, this looks like a good PR. Curious to see which test case breaks because of it.

arnog:
The PR is still a draft, and therefore cannot be merged. Let me know when it's ready for review.

samueltlg:
Good that this has not been merged yet! - detected a couple more related issues.
One of the commits addresses the failing numeric tests: in simple terms, not passing around empty 'metadata' permits usage of engine-scoped common symbols (-the reason for the failures being differing serialization, or number casting, between ce.One (et cetera) and equivalent 'ExactNumericValue's.)

src/compute-engine/boxed-expression/public.ts
```diff
@@ -2036,10 +2037,7 @@ export type EvaluateOptions = {
  * @category Boxed Expression
  */
 
-export type Metadata = {
-  latex?: string | undefined;
-  wikidata?: string | undefined;
-};
+export type Metadata = Pick<Attributes, 'latex' | 'wikidata'>;
```
samueltlg, line 2040: May wish to revert this: personally just thought it made sense.

src/compute-engine/compute-engine.ts
```diff
@@ -2015,10 +2015,13 @@ export class ComputeEngine implements IComputeEngine {
   _fn(
     name: MathJsonIdentifier,
     ops: ReadonlyArray<BoxedExpression>,
-    options?: Metadata & { canonical?: boolean }
```
samueltlg, line 2018: Quite sure that that was a mistake, or at least non-desirable

PR #214, opened 2024-11-01
fix: typo in NATIVE_JS_FUNCTIONS

### LINKED ISSUE #213: "Arcsin" can't be compiled
## Description

`ce.parse` parses the latex trigger `\\arcsin` to `"Arcsin"` (which is reasonable, of course), but `.compile()` does not know `"Arcsin"`.

### Steps to Reproduce

```
const expr = ce.parse('\\arcsin(x)');
const f = expr.compile();
```

### Actual Result

Throws error:
"Error: Unknown function Arcsin"

### Expected Result

The expression is compiled to a js function without throwing errors.

### Environment

Compute engine version: "^0.26.2"

rasmus-rasmus:
Closes #213

### DIFF
--- a/src/compute-engine/compile.ts
+++ b/src/compute-engine/compile.ts
@@ -100,7 +100,7 @@ const NATIVE_JS_FUNCTIONS: CompiledFunctions = {
     return `Math.acosh(1 / (${compile(x)}))`;
   },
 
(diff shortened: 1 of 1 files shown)


arnog:
Good catch. Thanks for the PR!

PR #215, opened 2024-11-08
Simplification Rules Test Cases and sgn function changes

andrew-murdza:
Simplification Rules Test Cases and sgn function changes

### DIFF
--- a/src/compute-engine/library/arithmetic.ts
+++ b/src/compute-engine/library/arithmetic.ts
@@ -56,7 +56,6 @@ import {
   canonicalPower,
   canonicalRoot,
   pow,
-  root,
 } from '../boxed-expression/arithmetic-power';
 import { parseType } from '../../common/type/parse';
 import { isSubtype } from '../../common/type/subtype';
@@ -184,9 +183,9 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
         if (ops.some((x) => x.isReal === false || x.isNaN)) return 'unsigned';
         if (ops.every((x) => x.is(0))) return 'zero';
         if (ops.every((x) => x.isNonNegative))
-          return ops.every((x) => x.isPositive) ? 'positive' : 'non-negative';
+          return ops.some((x) => x.isPositive) ? 'positive' : 'non-negative';
         if (ops.every((x) => x.isNonPositive))
-          return ops.every((x) => x.isNegative) ? 'negative' : 'non-positive';
+          return ops.some((x) => x.isNegative) ? 'negative' : 'non-positive';
         if (ops.every((x) => x.isReal)) return 'real';
         return undefined;
       },
@@ -210,7 +209,8 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
       sgn: ([x]) => {
         if (x.isLessEqual(-1)) return 'negative';
         if (x.isPositive) return 'positive';
-        if ((x.isNegative && x.isGreater(-1)) || x.is(0)) return 'zero';
+        if (x.isNonNegative) return 'non-negative';
+        if (x.isNonPositive && x.isGreater(-1)) return 'zero';
         if (x.isNonPositive) return 'non-positive';
         if (x.isReal == false && x.isNumberLiteral)
           return x.im! > 0 || x.im! <= -1 ? 'unsigned' : numberSgn(x.re); //.re and .im should be more general.
@@ -274,8 +274,8 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
         if (d.is(0) === false && n.is(0)) return 'zero';
         if (d.isPositive) return n.sgn;
         if (d.isNegative) return oppositeSgn(n.sgn);
-        if (n.is(0) && d.isFinite) return 'zero';
-        if (!n.is(0) && !d.is(0)) return 'real-not-zero';
+        if (n.is(0) || (n.isFinite && d.isInfinity)) return 'zero';
+        if (!n.is(0) && !d.is(0)) return 'not-zero';
         return undefined;
       },
 
@@ -404,7 +404,7 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
       sgn: ([x]) => {
         if (x.isNegative) return 'negative';
         if (x.isGreaterEqual(1)) return 'positive';
-        if ((x.isPositive && x.isLess(1)) || x.is(0)) return 'zero';
+        if (x.isNonNegative && x.isLess(1)) return 'zero';
         if (x.isNonNegative) return 'non-negative';
         if (x.isReal == false && x.isNumberLiteral)
           return x.im! < 0 || x.im! >= 1 ? 'unsigned' : numberSgn(x.re); //.re and .im should be more general.
@@ -641,7 +641,6 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
           });
           return sumNeg % 2 === 0 ? 'non-positive' : 'non-negative';
         }
-        if (ops.every((x) => !x.is(0) && x.isReal)) return 'real-not-zero';
         if (ops.every((x) => !x.is(0))) return 'not-zero';
         if (ops.every((x) => x.isReal)) return 'real';
         return undefined;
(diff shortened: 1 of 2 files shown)


test/compute-engine/simplify.test.ts
```diff
@@ -128,45 +128,6 @@ const CANONICALIZATION_TEST_CASES: TestCase[] = [
               //
             `,
   ],
-  ['x/(y/2)^3', '(8*x)/y^3'],
```
arnog, line 131: Why did those test cases were moved to RULE_TEST_CASE?

test/compute-engine/simplify.test.ts
```diff
@@ -937,53 +785,94 @@ const RULE_TEST_CASES: TestCase[] = [
 const RULES_USED: string[] = [];
 
 const RULES: Rule[] = [
-  '(-x)^n:even -> x^n',
-  '(-x)^n:odd -> -(x^n)',
+  /*@fixme
```
arnog, line 788: What is the problem with these rules? Are they not being run, or do they give an unexpected result, or...?

test/compute-engine/simplify.test.ts
```diff
@@ -1160,64 +960,31 @@ const RULES: Rule[] = [
...
   // {match:'\\infty^a',replace:'\\infty',condition:id=>id._a.isPositive===true},
 
-  //Logs and Infinity
-  '\\ln(\\infty) -> \\infty',
-  {
-    match: '\\log_c(\\infty)',
-    replace: '\\infty',
-    condition: ({ c }) => c.isGreater(1) === true,
-  },
+  //@fixme
```
arnog, line 982: What is the problem with this rule?

src/compute-engine/library/arithmetic.ts
```diff
@@ -830,8 +829,7 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
         const [base, exp] = args;
         return canonicalRoot(base, exp);
       },
-      evaluate: ([x, n], { numericApproximation }) =>
-        root(x, n, { numericApproximation }),
+      evaluate: ([x, n], { numericApproximation }) => x.root(n),
```
arnog, line 832: That doesn't sound correct: it ignores the numericApproximation flag. The previous implementation looks good.

src/compute-engine/library/arithmetic.ts
```diff
@@ -1395,7 +1392,7 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
         if (s === undefined) return undefined;
         if (['positive', 'zero'].includes(s)) return s;
         const n = reduceBigOp(ops[0], ops.slice(1), (acc, x) => acc + 1, 0);
-        if (s === 'real-not-zero' || s === 'negative')
+        if ((ops[0]?.type === 'real' && s === 'not-zero') || s === 'negative')
```
arnog, line 1395: Checking that type === 'real' is probably not correct (it could be 'integer' or 'rational' and this condition would presumably still apply). I think you meant to check ops[0]?.isReal, which will account for real or one of its subtype.

src/compute-engine/library/arithmetic.ts
```diff
@@ -184,9 +183,9 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
         if (ops.some((x) => x.isReal === false || x.isNaN)) return 'unsigned';
         if (ops.every((x) => x.is(0))) return 'zero';
         if (ops.every((x) => x.isNonNegative))
-          return ops.every((x) => x.isPositive) ? 'positive' : 'non-negative';
+          return ops.some((x) => x.isPositive) ? 'positive' : 'non-negative';
         if (ops.every((x) => x.isNonPositive))
-          return ops.every((x) => x.isNegative) ? 'negative' : 'non-positive';
+          return ops.some((x) => x.isNegative) ? 'negative' : 'non-positive';
         if (ops.every((x) => x.isReal)) return 'real';
```
arnog, line 189: I don't think this can ever be triggered, since there is a check for isReal === false at the beginning of the handler.

src/compute-engine/library/arithmetic.ts
```diff
@@ -641,7 +641,6 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
           });
           return sumNeg % 2 === 0 ? 'non-positive' : 'non-negative';
         }
-        if (ops.every((x) => !x.is(0) && x.isReal)) return 'real-not-zero';
         if (ops.every((x) => !x.is(0))) return 'not-zero';
         if (ops.every((x) => x.isReal)) return 'real';
```
arnog, line 645: There is a check for isReal at the beginning of the handler, so I don't think this is ever reached.

src/compute-engine/library/arithmetic.ts
```diff
@@ -641,7 +641,6 @@ export const ARITHMETIC_LIBRARY: IdentifierDefinitions[] = [
           });
           return sumNeg % 2 === 0 ? 'non-positive' : 'non-negative';
         }
-        if (ops.every((x) => !x.is(0) && x.isReal)) return 'real-not-zero';
         if (ops.every((x) => !x.is(0))) return 'not-zero';
```
arnog, line 644: There is a check for is(0) line 619, so I don't think this is ever reached. It also returns 'not-zero', while the check line 619 returns other values. Not sure which is correct.

arnog (commented):
Thanks for the PR. Could you clarify a few things in the comments I've left?

test/compute-engine/simplify.test.ts
andrew-murdza: It didn't need to be removed. I didn't realize t the fact that the RULE_TEST_CASE only includes stuff related to the rules because there are many cases there that don't use any of the rules

test/compute-engine/simplify.test.ts
andrew-murdza: They are not being run

test/compute-engine/simplify.test.ts
andrew-murdza: It does it by default. Neither or nor the default works for $e^\infty$ because it doesn't realize that $e>1$

src/compute-engine/library/arithmetic.ts
andrew-murdza: I don't remember making this change because I didn't look that the evaluate stuff. I believe this may be an issue with Github.

src/compute-engine/library/arithmetic.ts
andrew-murdza: It is my mistake for not checking possible values of type. I thought it was just real, imaginary, complex

src/compute-engine/library/arithmetic.ts
andrew-murdza: You're right but your top line is wrong. If some x is not real it should return undefined not unsigned (consider i+(-i)) so you need to remove that. So you would do a test for x.isNaN to return unsigned. This means that the test for isReal is no longer there which is why the other line is needed.

src/compute-engine/library/arithmetic.ts
andrew-murdza: I agree with your point.

src/compute-engine/library/arithmetic.ts
andrew-murdza: This is a problem with .is(0). It should return true, false, or undefined. If you do that (which I think I assumed it did) then !x.is(0) and x.is(0) can both be false. Otherwise, x^2(2) will be considered not 0 since (x^2).is(0) returns false.

test/compute-engine/simplify.test.ts
arnog: OK, I'll move them back. The cases that don't use any of the rules should also be moved out of RULE_TEST_CASE but that can be done later.

test/compute-engine/simplify.test.ts
arnog: OK

src/compute-engine/library/arithmetic.ts
arnog: I don't think it's "my" top line, but if it's incorrect it should be fixed.

src/compute-engine/library/arithmetic.ts
arnog: `x.is(0)` is either true or false. Something either *is* 0, or it is not. Perhaps you meant to check that `x.sgn` was `zero` or not?

andrew-murdza: I see. Those checks should be using x.sgn not is(0).

andrew-murdza: is(0) works for showing something is zero but !is(0) cannot be used to check if something is non-zero (only if it is not identically 0).

PR #218, opened 2024-11-23
Optimize parsing of invisible operator operands

dcbradley:
Some expressions take a very long time to parse, scaling exponentially with the length of the expression.  The following example takes over 20 minutes to parse without this patch and less than a second with the patch.

    xxxxxxxxxxxxxxxx\left[yyyyyyyy\left[zzzzzzzzzz\left[\left[\left[x.x\right]\right]\right]\right]\right]

It is an invalid expression.  If that could be determined in general without attempting to parse it, perhaps the parsing efficiency wouldn't matter, but I don't know of a way to achieve that.

The cause of the blow-up in execution time is a combination of backtracking to the opening braces due to the invalid part of the expression and continued attempts to parse forward again due to the loop over invisible operator operands at every level in the recursive chain of parseExpression() calls.

This patch avoids many redundant attempts to parse the next invisible operator operand by only doing that in the parent parseExpression() and not in its immediate children.  This is implemented in a somewhat ad-hoc way rather than splitting the multiplication precedence level in two.  If you would like, I'd be happy to submit a patch that does the latter.  I think that will disturb more code, so that is why I did it this way.

Thanks for listening, and thanks for the wonderful software!

### DIFF
--- a/src/compute-engine/latex-syntax/dictionary/definitions-arithmetic.ts
+++ b/src/compute-engine/latex-syntax/dictionary/definitions-arithmetic.ts
@@ -18,6 +18,7 @@ import {
   Serializer,
   Parser,
   LatexDictionary,
+  INVISIBLE_OP_PRECEDENCE,
   MULTIPLICATION_PRECEDENCE,
   ADDITION_PRECEDENCE,
   ARROW_PRECEDENCE,
@@ -1153,7 +1154,7 @@ export const DEFINITIONS_ARITHMETIC: LatexDictionary = [
     latexTrigger: ['\\sum'],
     precedence: ADDITION_PRECEDENCE,
     name: 'Sum',
-    parse: parseBigOp('Sum', ADDITION_PRECEDENCE),
+    parse: parseBigOp('Sum', MULTIPLICATION_PRECEDENCE),
     serialize: serializeBigOp('\\sum'),
   },
   {
@@ -1265,7 +1266,7 @@ function getIndexes(
     .filter((x) => x !== undefined);
 }
 
-function parseBigOp(name: string, prec: number) {
+function parseBigOp(name: string, minPrec: number) {
   return (parser: Parser): Expression | null => {
     parser.skipSpace();
 
@@ -1292,7 +1293,7 @@ function parseBigOp(name: string, prec: number) {
     for (const indexinSet of indexes)
       parser.addSymbol(indexinSet.index, 'symbol');
 
-    const fn = parser.parseExpression({ minPrec: prec + 1 });
+    const fn = parser.parseExpression({ minPrec: minPrec });
 
     parser.popSymbolTable();
 
--- a/src/compute-engine/latex-syntax/parse.ts
+++ b/src/compute-engine/latex-syntax/parse.ts
@@ -19,6 +19,7 @@ import {
   Delimiter,
   Terminator,
   Parser,
+  INVISIBLE_OP_PRECEDENCE,
   MULTIPLICATION_PRECEDENCE,
   SymbolTable,
   SymbolType,
@@ -1356,7 +1357,7 @@ export class _Parser implements Parser {
 
       // No group, but arguments without parentheses are allowed
       // Read a primary
-      const primary = this.parseExpression({ ...until, minPrec: 390 });
+      const primary = this.parseExpression({ ...until, minPrec: MULTIPLICATION_PRECEDENCE });
       return primary === null ? null : [primary];
     }
 
@@ -1987,13 +1988,13 @@ export class _Parser implements Parser {
         this.skipSpace();
 
         let result = this.parseInfixOperator(lhs, until);
-        if (result === null) {
+        if (result === null && until.minPrec <= INVISIBLE_OP_PRECEDENCE) {
           // If any operator, no sequence to apply
           if (this.peekDefinitions('operator').length === 0) {
             // No infix operator, join the expressions with a Sequence
             const rhs = this.parseExpression({
               ...until,
-              minPrec: MULTIPLICATION_PRECEDENCE,
+              minPrec: INVISIBLE_OP_PRECEDENCE+1,
             });
             if (rhs !== null) {
               if (operator(lhs) === 'InvisibleOperator') {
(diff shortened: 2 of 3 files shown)


arnog:
Thanks, that's a great catch. Surprised parsing could take 20 mins, but I guess it's the power of exponential recursion!

Your patch works and is OK. As a stylistic thing, I prefer to avoid having boolean arguments, because they're not self-documenting, and it's easy to forget if you should pass true or false. Instead, I would rather have a named property in a structure. Not super important in this case, since the argument is only used in a recursive call of the function.

Nonetheless, I think I would rather add a `invisibleOperator` to the `until` (i.e. `Terminator`) structure, unless you want to go ahead and split the multiplication precedence (I assume that means there would be a "regular" one and one for "invisible" multiplication?).

Let me know how you want to proceed. If you'd like to explore either of those options, great, otherwise I'll pickup your patch and tweak it. Either way, thanks again for submitting it!

dcbradley:
As an experiment, I have redone the patch by using `until.minPrec` itself to implement the optimization.  A new `INVISIBLE_OP_PRECEDENCE` that is less than `MULTIPLICATION_PRECEDENCE` was added.

In hindsight, it would be even simpler to change the invisible operator parsing loop to use `minPrec=MULTIPLICATION_PRECEDENCE+1` when parsing arguments.  Then there is no need to introduce `INVISIBLE_OP_PRECEDENCE`.  The two different approaches produce semantically equivalent results when invisible operator is interpreted as multiplication, but are there other cases that need to be considered?

In the current patch, and prior to this patch, with canonical=false, `ab*cd` becomes:
    ["InvisibleOperator","a",["Multiply","b","c"],"d"]

If the alternative approach I mentioned above were used, or if `INVISIBLE_OP_PRECEDENCE` is set equal to MULTIPLICATION_PRECEDENCE, the result would instead be:
    ["InvisibleOperator",["Multiply",["InvisibleOperator","a","b"],"c"],"d"]

Whichever choice is made, this general approach introduces enforcement of precedence for the invisible operator, which causes important semantic changes.  Previously, `parseExpression()` did not check `until.minPrec` before entering the invisible operator loop.  This means, prior to this patch, `1/2a` got parsed like `1/(2a)`, whereas now it is like `(1/2)a`.  The new behavior is consistent with `1/2*a` being interpreted like `(1/2)*a` irrespective of this patch.  The latter is consistent with the defined precedence of division being higher than multiplication, which is at first surprising, since it is not like a calculator or programming language, but probably more consistent with mathematical tradition.

The power operator is implemented with a custom parser, so I don't think it is impacted.

The negation operator is defined with higher order of precedence than multiplication, so it is impacted.  I'm not aware of any case where that would create a semantic difference, but it certainly does impact the parsed forms of expressions such as `-ab`, which is now like `(-a)b` rather than `-(ab)`.  All of the tests that fail are due to this change.

The `\prod` operator is impacted.  Previously, it was parsing its argument with `minPrec=MULTIPLICATION_PRECEDENCE+1`.  Therefore, `\prod_{n=1}^3 n*m` was interpreted the same as `(\prod_{n=1}^3 n)*m`, whereas `\prod_{n=1}^3 nm` was interpreted the same as `\prod_{n=1}^3 (nm)`.  I believe the latter is intended, so this patch makes the visible multiplication case work this way as well.

The `\sum` operator is impacted as a side-effect of a fix to `parseBigOp()` for `\prod`.  The `prec` argument to `parseBigOp()` was being treated as one less than the minimum precedence.  To fix `\prod`, this patch changes the argument to `minPrec`, which is treated as the minimum precedence.  I had to choose a new precision to pass into `parseBigOp()` for `\sum`.  It could have been changed to `ADDITION_PRECEDENCE+1` to avoid any change in behavior, but I am guessing that `INVISIBLE_OP_PRECEDENCE` is what is actually desired.  The precedence levels lower than that do not appear to be things that would be expected to be slurped into the argument of `\sum`.  I mention it in case you disagree.

dcbradley:
Since I was wrong that setting `INVISIBLE_OP_PRECEDENCE` less than `MULTIPLICATION_PRECEDENCE` would preserve the old order of operations between multiplication and invisible operator, I don't see any reason to stick with that.  It seems to me it should either be equal or perhaps greater.  None of these choices conflict with the intended side-effect of fixing the invisible operator performance problem, because in all cases, the value of `minPrec` used when parsing the invisible operator operands would be greater than `INVISIBLE_OP_PRECEDENCE`, which would prevent the recursive chain of `parseExpression()` calls from all hammering on a failed expression like they were.

An advantage of making the precedence equal is that `INVISIBLE_OP_PRECEDENCE` could just go away.  However, I am wondering if there are branches of math where one might expect the invisible operator to have higher precedence than multiplication type operators.

arnog:
In general, and for example when using TI calculators, the invisible operator is considered higher precedence than multiplication or division, so that `1/2a` is parsed as `1/(2*a)`.

arnog:
That looks good to me. The differences are either not very important (negation) or actually an improvement.

Thanks for looking into this and being so thorough.

dcbradley:
I redid the patch.  There are two things worth noting.  One involves mixed numbers:
    2\\frac{5}{7}\\times\\frac{7}{9}

The above expression used to parse (non-canonically) to this:
    ["InvisibleOperator",2,["Multiply",["Divide",5,7],["Divide",7,9]]]

Now it parses (non-canonically) to this:
    ["Multiply",["InvisibleOperator",2,["Divide",5,7]],["Divide",7,9]]

The canonical parser interprets the invisible operator in the second case as a mixed number `2+5/7`, whereas in the previous case, it interprets the invisible operator as multiplication.  I think this semantic change is reasonable, but since this case was in the test suite, perhaps you had a reason to prefer the previous behavior.

The second thing of note is that my previous assumption about the relative precedence of negation and exponentiation was wrong.  It doesn't work like this:
    -a^bc = ((-a)^b)c

Instead, it works like this:
    -a^bc = (-(a^b))c

As far as I can tell, this is intended, so all is well.

Other than those two things, I didn't find any surprises.

dcbradley:
If you would like a second commit with changes to the test suite, I'd be happy to do that.  I didn't want to combine that with the code change, so you could examine the tests that fail and agree that those changes in behavior are all fine.

arnog:
That looks good, thank you. I'll have a look at the mixed fraction case. It seems reasonable to me, but I've gotten some feedback previously on it, which may be why there was this test case. Either way, I'm sure we can make it work. I'll need before I can merge it, but the ball is in my court now.

Thanks again for this great contribution.

arnog:
`2\\frac{5}{7}\\times\\frac{7}{9}` was parsed incorrectly previously, this PR fixes it.
All good, updated the tests.

PR #226, opened 2024-12-23
Draft: (partial) fix for 'ce.declare' with SymbolDefinition using prop. 'type'

samueltlg:
Noticed that `ce.declare` errors when supplying a _SymbolDefinition_ of form `{type: TypeString}` for second arg.

Current changes appear to address the issue, but appears to surface or reveal a few existing test errs. where throughout these, type values given through 'SymbolDefinition.type' do not conform to current type definitions: e.g. `{ type: 'set<integers>', ... }` is given whereas in this context, according to definition of the type '*Type*', a `{kind: 'set', elements: 'integers' }` is expected instead. Did try to investigate further but opened a bit of a can of worms.

Believe that almost all errs. can be addressed fairly easily with present changes, but may require some alterations.
For reference, for the commit at which this was branched, 4 'snapshot' errors were already present split between tests '*compute-engine/compile.test.ts*', and '*compute-engine/compile.test.ts*'. This change surfaces ~18 newer

### DIFF
--- a/src/compute-engine/compute-engine.ts
+++ b/src/compute-engine/compute-engine.ts
@@ -1371,8 +1371,10 @@ export class ComputeEngine implements IComputeEngine {
     if (
       typeof def === 'object' &&
       'type' in def &&
-      ((typeof def.type === 'string' && isSubtype(def.type, 'function')) ||
-        (def.type && isValidType(def.type)))
+      !(
+        (typeof def.type === 'string' && isSubtype(def.type, 'function')) ||
+        (def.type && isValidType(def.type))
+      )
     ) {
       throw new Error(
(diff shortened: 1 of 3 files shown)


src/compute-engine/compute-engine.ts
```diff
@@ -1371,8 +1371,10 @@ export class ComputeEngine implements IComputeEngine {
     if (
       typeof def === 'object' &&
       'type' in def &&
-      ((typeof def.type === 'string' && isSubtype(def.type, 'function')) ||
-        (def.type && isValidType(def.type)))
+      !(
+        (typeof def.type === 'string' && isSubtype(def.type, 'function')) ||
+        (def.type && isValidType(def.type))
+      )
```
samueltlg, line 1377: Think this was a bug/typo? I.e. final cond. to be negated.

Issue #232, opened 2025-03-19
Error when importing package in a CommonJS module


yongla:
## Description

The CommonJS files of the build are treated as ES module files resulting in an error when importing this package in a CommonJS module.

I believe this is due to the way this package has set up its dual CJS/ESM builds and exports. There is a helpful tool called publint that highlights why this is happening for this project: https://publint.dev/@cortex-js/compute-engine@0.28.0

### Steps to Reproduce

1. Have a CommonJS file that imports `@cortex-js/compute-engine` (v0.28.0)
2. Run the code that tries to import the package
3. Notice an error is thrown as shared below

I've also created a basic sandbox showcasing the error: https://codesandbox.io/p/devbox/compute-engine-in-commonjs-mkpsl5

### Actual Result

An error like this is thrown at runtime when trying to import the package:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /project/sandbox/node_modules/@cortex-js/compute-engine/dist/compute-engine.min.js from /project/sandbox/index.js not supported.
compute-engine.min.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename compute-engine.min.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /project/sandbox/node_modules/@cortex-js/compute-engine/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

### Expected Result

No error when importing the package in a CommonJS module

### Environment

This is the first time I'm using this package, so I'm not sure if this used to work differently in previous version. I've tested this in an isolated node environment running v21.7.3 (the code sandbox above) and in the browser as well within my application.

**Compute Engine version**  0.28.0

arnog:
Fixed

PR #233, opened 2025-03-19
Fix CommonJS build and exports

yongla:
This is a fix for [this issue](https://github.com/cortex-js/compute-engine/issues/232) I created.

## Context 
With the way this package exports its CommonJS build, CommonJS modules importing `@cortex-js/compute-engine` treat the CommonJS files it exports as an ES module instead, causing an error.

The tool publint summarizes the issue nicely: https://publint.dev/@cortex-js/compute-engine@0.28.0

Example error message:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /project/sandbox/node_modules/@cortex-js/compute-engine/dist/compute-engine.min.js from /project/sandbox/index.js not supported.
compute-engine.min.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename compute-engine.min.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /project/sandbox/node_modules/@cortex-js/compute-engine/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

## Proposed fix

The proposed fix here is to have the CommonJS output of the build use the `.cjs` extension, and have the package's `exports` reflect that as well. This is also what publint recommends.

## Testing

I've tested importing a local build of this project with the code changes here and verified the import error goes away in both node/browser environments, but I haven't done any other particular testing besides that.


arnog:
Thanks for filing the issue and for the fix. Bundling is a bit like magic to me, so hard to say if this will address the issue, but I think most people using the library used ESM rather than CJS, so it might have been broken for a while... Let me know after I publish a version with the fix if it does not address your issue as expected.

yongla:
Thanks for the quick review! And sounds good, I'll definitely reach back out if I still have issues 👍

Issue #231, opened 2025-02-24
Inconsistencies with handling small fractions


Victorsss-Orz:
### Steps to Reproduce

Evaluating 10^{-14}, 10^{-14}/10, 10^{-14}/2, and 10^{-14}/1.75. 

### Actual Result

Evaluating 10^{-14} gets 0, where as evaluating 10^{-14}/10 gives 10^{-15}. Moreover, evaluating 10^{-14}/2 gives the correct 1/20000..., but10^{-14}/1.75 gives 0. 

### Expected Result

The values should all be non-zero

### Environment

> Is this a regression: Both versions I have are not working

**Compute Engine version**  0.27.0 and 0.28.0

Victorsss-Orz:
This is reported in a calculator UI I've built with this library. https://github.com/PrairieLearn/PrairieLearn/issues/11426 Item 3

nwalters512:
A more concerning example is `10^{-14}/(1.5 * 10^{-15})`, which evaluates to `NaN`. However, other similar things work:

- `10^{-14}/(1 * 10^{-15})` evaluates to 10
- `10^{-14}/(2 * 10^{-15})` evaluates to 5

In general, any integer coefficient in the denominator produces the correct value, whereas any non-integer coefficient results in `NaN`. It seems like `10^{-15}` is being "converted" to 0 too eagerly.

Extra2001:
I also encountered the same issue after updating ComputeEngine to 0.27.0 and 0.28.0 which introduced `ce.tolerance`. When I calculate the simple `10 ^ {-11}`, I get `0`. But strangely, setting `ce.tolerance = 1e-50` doesn't work for me.

Extra2001:
As I read and tested the source code, I found that the author only compared the numbers with the default tolerance in the `chop()` function, not the `tolerance` specified by the user when creating `ce` This is the direct cause of these errors.

`chop()` is in [big-numeric-value.ts](https://github.com/cortex-js/compute-engine/blob/main/src/compute-engine/numeric-value/big-numeric-value.ts). As follows: 

```ts
function chop(n: number): number {
  if (Math.abs(n) <= DEFAULT_TOLERANCE) return 0;

  return n;
}
```

Here, the `DEFAULT_TOLERANCE` is a constant value defined as follow: 

```ts
// Mathematica has a default tolerance of 10^-10
// Numpy has a default absolute tolerance of 1e-8 (1e-5 for relative)
[...]

PR #236, opened 2025-03-30
fix: use the tolerance defined by ce in numeric chop

Extra2001:
This will close the issue #231.
While changing `DEFAULT_TOLERANCE` fixes the problem, `chop` in the calculation are still worrisome.

### DIFF
--- a/src/compute-engine/boxed-expression/terms.ts
+++ b/src/compute-engine/boxed-expression/terms.ts
@@ -185,8 +185,8 @@ function nvSum(
   const makeExact = (x) => new ExactNumericValue(x, factory, bignum);
   const factory =
     ce.precision > MACHINE_PRECISION
-      ? (x) => new BigNumericValue(x, bignum)
-      : (x) => new MachineNumericValue(x, makeExact);
+      ? (x) => new BigNumericValue(x, bignum, ce.tolerance)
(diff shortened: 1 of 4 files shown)


arnog:
Thanks for the PR.
I think you're right that the issue is that a `chop()` was incorrectly applied when doing some calculations. I have submitted a fix that removes it and I think that's a better way to address the issue.
I don't think that tying the tolerance of the `BigNumericValue` to the tolerance of the compute engine is right: the `chop()` in `BigNumericValue` is only applied when converting to a machine number, and the tolerance/precision of a machine number is fixed, and not dependent on the precision/tolerance of the compute engine. It probably should not be using `DEFAULT_TOLERANCE`, as this is confusing, but its own tolerance value.

Extra2001:
Thank you for your reply! I see your commit solves this issue. But there are still some problems: if the input number contains machine number form, if it is less than `1e-10`, it will still be parsed as `0`, and the json obtained by parsing will also get `0`. Is this the expected parse result?

Test input Latex: ` \frac{10^{-14}}{1e-15} `
Output MathJSON: `["Divide",["Power",10,["Negate",14]],0]`

arnog:
Good point, that is indeed the problem in `BigNumericValue` and `MachineNumericValue`. There were some extraneous `chop()` in there, and I've also adjusted them to use the tolerance of machine numbers, which is the intent.

PR #238, opened 2025-04-07
Feat/fix: revised canonical forms; tests; preliminary & associated fixes

samueltlg:
Still more work & tests incoming - which is done but needs verification & review - for CanonicalForms `Multiply`, `Divide` & `Number` (tests). `Power` form most notably up for now: this being the trickiest on

If you review this set of changes, will make any requested changes along with the next/remaining batch of work: this should be next Wed./Thu. evening.

The individual commit messages are good sources of info. WRT changes.

Have a batch of incoming code-comments to make (will do tomorrow) which will likely clarify, and sidestep confusion, and highlight key-points/requests.

Some outstanding queries:
- WRT *symbols* being involved in canonicalization (notably, 'Power'):
  - Only recently came to mind that the `holdUntil` symbol attribute is not accounted for during the check on operand values (which could potentially be symbols): should this be the case?
    - Yet furthermore, if this is to be checked, i.e. that the 'holdUntil' value of a symbol-definition is `never`, then would it not be the case that symbols are substituted with values, before canonicalization, either partial or full, takes place? That being said, it would be the case that accounting for symbol values during application of canonical-forms, is unnecessary, since they would be substituted beforehand anyway (and any existing symbols would therefore have a 'holdUntil' value of 'evaluate', 'N', etc...)
   - With the above in mind, in the case of Power-form, for example, should operations such as `1^x`, where `x` is declared as a constant of value `1`, *ever* take place... ?

### DIFF
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -661,10 +661,6 @@ export abstract class _BoxedExpression implements BoxedExpression {
     throw new Error(`Can't change the value of \\(${this.latex}\\)`);
   }
 
-  get constantValue(): number | boolean | string | object | undefined {
-    return undefined;
-  }
-
   get type(): BoxedType {
     return BoxedType.unknown;
   }
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -224,7 +224,7 @@ export abstract class _BoxedExpression implements BoxedExpression {
   toLatex(options?: Partial<SerializeLatexOptions>): LatexString {
     // We want to use toMathJson(), not .json, so that we have all
     // the digits for numbers, repeated decimals
-    const json = this.toMathJson();
+    const json = this.toMathJson({ prettify: options?.prettify ?? true });
 
     let effectiveOptions: SerializeLatexOptions = {
       imaginaryUnit: '\\imaginaryI',
@@ -249,7 +249,7 @@ export abstract class _BoxedExpression implements BoxedExpression {
       notation: 'auto',
       avoidExponentsInRange: [-7, 20],
 
-      prettify: true,
+      prettify: true, // (overridden subseq. by options)
 
       invisibleMultiply: '', // '\\cdot',
       invisiblePlus: '', // '+',
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -419,10 +419,6 @@ export abstract class _BoxedExpression implements BoxedExpression {
     return false;
   }
 
-  /** Literals (number, string, boolean) are constants. Some symbols
-   * may also be constants (e.g. Pi, E, True, False). Expressions of constant
-   * symbols are also constants (if the function is pure).
-   */
   get isConstant(): boolean {
     return true;
   }
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -665,6 +665,10 @@ export abstract class _BoxedExpression implements BoxedExpression {
     throw new Error(`Can't change the value of \\(${this.latex}\\)`);
   }
 
+  get constantValue(): number | boolean | string | object | undefined {
+    return undefined;
+  }
+
   get type(): BoxedType {
     return BoxedType.unknown;
   }
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -134,7 +134,7 @@ export abstract class _BoxedExpression implements BoxedExpression {
     if (this.isInfinity) {
       if (this.isPositive) return Infinity;
       if (this.isNegative) return -Infinity;
-      return NaN;
+      return '~oo'; // ComplexInfinity
     }
     if (typeof this.string === 'string') return this.string;
     if (typeof this.symbol === 'string') return this.symbol;
--- a/src/compute-engine/boxed-expression/arithmetic-power.ts
+++ b/src/compute-engine/boxed-expression/arithmetic-power.ts
@@ -5,6 +5,7 @@ import type { Rational } from '../numerics/types';
 import { canonicalAngle, getImaginaryFactor } from './utils';
 import { apply, apply2 } from './apply';
 import { SMALL_INTEGER } from '../numerics/numeric';
+import type { NumericType, Type } from '../types';
 
 function isSqrt(expr: BoxedExpression): boolean {
   return (
@@ -31,38 +32,182 @@ export function asRadical(expr: BoxedExpression): Rational | null {
   return null;
 }
 
+/**
+ *
+ * Produce the canonical form of the operands of a Power expression, returning either the operation
+ * result (e.g. 'a^1 -> a'), an alternate expr. representation ('a^{1/2} -> Sqrt(a)'), or an
+ * unchanged 'Power' expression. Operations include:
+ * 
+ * - @todo
+ * 
+ * Both the given base and exponent can either be canonical or non-canonical: with fully
+ * canonicalized args. lending to more simplifications.
+ * 
+ * Returns a canonical expr. is both operands are canonical.
+ 
+ * @export
+ * @param a
+ * @param b
+ * @returns
+ */
 export function canonicalPower(
   a: BoxedExpression,
   b: BoxedExpression
 ): BoxedExpression {
   const ce = a.engine;
-  a = a.canonical;
-  b = b.canonical;
-  if (a.is(0)) {
+
+  const fullyCanonical = a.isCanonical && b.isCanonical;
+  const unchanged = () =>
+    ce._fn('Power', [a, b], { canonical: fullyCanonical });
+
+  // Exclude cases - which may otherwise be valid - of the exponent either: being a function (e.g.
+  // '0 + 0'), a non-constant value (symbols with 'assumed' values, impure functions), or a
+  // non-numeric type (no concern for this).
+  //
+  // @consider:possible exceptions where function-expressions are reasonable :Rational,Half,
+  // Negate... (However, provided that canonicalNumber provided prior, should not be missing anything
+  // here)
+  if (
+    b.isFunctionExpression ||
+    !b.isConstant || //(!notably, should exclude symbol assumptions)
+    !b.type.matches('number' as Type)
+  )
+    return unchanged();
+
+  // 'a' is constant and of type 'number'. Used often.
+  // (note that this could still be a numeric expression, too...)
+  const aIsNum = a.isConstant && a.type.matches('number' as NumericType);
+
+  // Zero as base
+  if (aIsNum && a.is(0)) {
+    if (b.type.matches('imaginary' as NumericType) || b.isNaN) return ce.NaN;
+
     if (b.is(0)) return ce.NaN;
-    if (b.isPositive) return ce.Zero;
-    if (b.isNegative) return ce.ComplexInfinity;
+
+    if (b.isInfinity) {
+      // 0^∞ = 0 (because for all complex numbers z near 0, z^∞ -> 0).
+      if (b.isPositive) return ce.Zero; // 0^∞ = 0
+      // 0^-∞ = ~∞
+      if (b.isNegative) return ce.ComplexInfinity;
+      return ce.NaN; // 0^~∞ = NaN
+    }
+    //(note: these should be applicable only to the reals)
+    if (b.isGreater(0)) return ce.Zero;
+    if (b.isLess(0)) return ce.ComplexInfinity;
+
+    return unchanged(); // No other canonicalization cases with this base
   }
-  if (a.is(1) || b.is(0)) return ce.One;
+
+  // Zero as exponent
+  if (b.is(0)) {
+    if (aIsNum) return a.isFinite ? ce.One : ce.NaN;
+    return unchanged();
+  }
+
+  // One as base
+  // (note: 1^∞ = NaN - Because there are various cases where lim(x(t),t)=1, lim(y(t),t)=∞ (or -∞),
+  // but lim( x(t)^y(t), t) != 1.)
+  if (aIsNum && a.is(1)) return b.isFinite ? ce.One : ce.NaN;
+
+  // One as exponent
   if (b.is(1)) return a;
+
+  // -1 exponent
+  if (b.is(-1)) {
+    if (aIsNum) {
+      // (-∞)^-1 = 0, ∞^-1 = 0  (exclude ~oo)
+      if (a.isInfinity && (a.isNegative || a.isPositive)) return ce.Zero;
+
+      // (-1)^-1 = -1
+      if (a.is(-1)) return ce.NegativeOne;
+
+      // 1^-1 = 1
+      if (a.is(1)) return ce.One;
+    }
+
+    // (note: case of `0^-1 = ~∞` is covered prior...)
+    return a.inv();
+  }
+
+  //Infinity exponents
+  if (b.isInfinity && aIsNum) {
+    // x^oo
+    if (b.isPositive) {
+      // (note: 0^∞ = 0, 1^∞ = NaN, covered prior)
+
+      // (-1)^∞ = NaN
+      // Because of oscillations in the limit.
+      if (a.is(-1)) return ce.NaN;
+
+      //↓note:the case for all infinites.
+      if (a.isInfinity) return ce.ComplexInfinity;
+
+      if (a.isNaN) return ce.NaN;
+
+      //↓numeric-expr. bases included: e.g. '{2+3}^oo'
+      if (a.isReal) {
+        if (a.isGreater(1)) return ce.PositiveInfinity;
+        if (a.isLess(-1)) return ce.ComplexInfinity;
+        // Must be '-1 < a < 1', excluding zero
+        return ce.Zero;
+      }
+
+      return unchanged();
+    }
+
+    // x^-oo
+    if (b.isNegative) {
+      if (a.is(-1)) return ce.NaN;
+      //Same result for all infinity types...
+      if (a.isInfinity) return ce.Zero;
+
+      if (a.isNaN) return ce.NaN;
+
+      if (a.isReal) {
+        if (a.isGreater(0)) return a.isLess(1) ? ce.PositiveInfinity : ce.Zero;
+        // Must be < 0
+        return a.isGreater(-1) ? ce.ComplexInfinity : ce.Zero;
+      }
+      return unchanged();
+    }
+
+    //Must be 'x^ComplexInfinity'
+    // b^~∞ = NaN
+    // Because b^z has no limit as z -> ~∞.
+    return ce.NaN;
+  }
+
+  //'Infinity^Complex'
+  if (a.isInfinity) {
+    // If the exponent is pure imaginary, the result is NaN
+    //(↓fix?:ensure both these cases narrow down to 'b' being a num./symbol literal)
+    if (b.type.matches('imaginary' as NumericType)) return ce.NaN;
+    if (b.type.matches('complex') && !isNaN(b.re)) {
+      if (b.re > 0) return ce.ComplexInfinity;
+      if (b.re < 0) return ce.Zero;
+    }
+  }
+
+  // Fractional exponents
+  //---------------------
   if (b.is(0.5)) return canonicalRoot(a, 2);
-
   const r = asRational(b);
-  if (r !== undefined && r[0] === 1) return canonicalRoot(a, ce.number(r[1]));
 
-  return ce._fn('Power', [a, b]);
+  //1/3, 1/4...
+  if (r !== undefined && r[0] === 1 && r[1] !== 1)
+    return canonicalRoot(a, ce.number(r[1]));
+
+  return unchanged();
 }
 
 export function canonicalRoot(
   a: BoxedExpression,
   b: BoxedExpression | number
 ): BoxedExpression {
-  a = a.canonical;
   const ce = a.engine;
   let exp: number | undefined = undefined;
   if (typeof b === 'number') exp = b;
   else {
-    b = b.canonical;
     if (b.isNumberLiteral && b.im === 0) exp = b.re;
   }
 
--- a/src/compute-engine/boxed-expression/box.ts
+++ b/src/compute-engine/boxed-expression/box.ts
@@ -191,17 +191,15 @@ export function boxFunction(
   //
   if (name === 'Number' && ops.length === 1) return box(ce, ops[0], options);
 
-  const canonicalNumber =
-    structural === false &&
-    (options.canonical === true ||
-      options.canonical === 'Number' ||
-      (Array.isArray(options.canonical) &&
-        options.canonical.includes('Number')));
+  const canonicalNumber = structural === false && options.canonical === true;
 
+  // If canonical, handle cases of various expression structures being able to be cast as
+  // BoxedNumbers (some cases of Negate, Rational, Divide, Complex), or 'de-number' some 'borderline
+  // invalid' boxed number-like expressions
+  // (!@note: this procedure is similarly repeated within the 'number' CanonicalForm, but the
+  // numberForm variant more simply applies to fully BoxedExprs., and during partial canonicalization
+  // only)
   if (canonicalNumber) {
-    // If we have a full canonical form or a canonical form for numbers
-    // do some additional simplifications
-
     //
     // Rational (as Divide)
     //
--- a/src/compute-engine/boxed-expression/boxed-function.ts
+++ b/src/compute-engine/boxed-expression/boxed-function.ts
@@ -197,10 +197,8 @@ export class BoxedFunction extends _BoxedExpression {
 
   get isPure(): boolean {
     if (this._isPure !== undefined) return this._isPure;
-    if (!this.isCanonical) {
-      this._isPure = false;
-      return false;
-    }
+    if (this.isCanonical === false) return this.canonical.isPure;
+
     let pure = this.functionDefinition?.pure ?? false;
 
     // The function might be pure. Let's check that all its arguments are pure.
@@ -210,9 +208,6 @@ export class BoxedFunction extends _BoxedExpression {
     return pure;
   }
 
-  /** The value of the function is constant if the function is
-   * pure, and all its arguments are constant.
-   */
   get isConstant(): boolean {
     if (!this.isPure) return false;
     return this._ops.every((x) => x.isConstant);
@@ -220,6 +215,15 @@ export class BoxedFunction extends _BoxedExpression {
 
   get constantValue(): number | boolean | string | object | undefined {
     return this.isConstant ? this.value : undefined;
+  }
+
+  /**
+   *
+   * @inheritdoc
+   */
+  get value(): number | boolean | string | object | undefined {
+    if (!this.isPure) return undefined;
+    return this.N().valueOf();
   }
 
   get json(): Expression {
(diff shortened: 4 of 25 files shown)


test/utils.ts
```diff
@@ -97,15 +99,41 @@ export function simplify(latex: string): string {
   return exprToString(engine.parse(latex)?.simplify());
 }
 
-export function checkJson(inExpr: SemiBoxedExpression | null): string {
+type ExprVariant =
```
samueltlg, line 102: I ended up not using this (as implemented in 'checkJson') in the end, due to using a custom solution in *canonical-form.test.ts* which prints a 'canonical-forms' variant: but conceivably, could see myself as having a use for this.
If not your cup of tea, do do away!

test/utils.ts
```diff
@@ -97,15 +99,41 @@ export function simplify(latex: string): string {
...
+): string {
   if (!inExpr) return 'null';
   try {
-    const precision = engine.precision;
-    engine.precision = 'auto';
-
-    const boxed = exprToString(engine.box(inExpr, { canonical: false }));
-
-    const expr = engine.box(inExpr);
+    const ce =
```
samueltlg, line 117: As with one or two other locations within this file: this makes sense if wishing to use these utilities for expr's not parsed with the default (this file) test engine?

src/compute-engine/boxed-expression/boxed-symbol.ts
```diff
@@ -437,7 +440,8 @@ export class BoxedSymbol extends _BoxedExpression {
     // Widen the type, if it was previously inferred
     if (
       def instanceof _BoxedSymbolDefinition &&
-      (def.inferredType || def.type.isUnknown)
+      (def.inferredType || def.type.isUnknown) &&
+      !def.isConstant //'constant' symbols may still be inferred (i.e. value given but no type)
```
samueltlg, line 444: Another possibility may be raising an exception, if attempt to *infer* for a constant-symbol ?

src/compute-engine/global-types.ts
```diff
@@ -1048,12 +1048,24 @@ export interface BoxedExpression {
    *
    * Infer the type of this expression.
    *
-   * If the type of this expression is already known, return `false`.
+   * Effective only for overall BoxedExpression types which are *non-constant* and therefore for
```
samueltlg, line 1054: Do please verify, or amend.

samueltlg: `for functions, narrows the *(return) type*`, may be incorrect

src/compute-engine/boxed-expression/boxed-symbol.ts
```diff
@@ -414,10 +414,13 @@ export class BoxedSymbol extends _BoxedExpression {
   }
 
   /**
+   *
    * Subsequent inferences will narrow the domain of the symbol.
```
samueltlg, line 418: ... will *narrow*: this is outdated now ?

src/compute-engine/boxed-expression/boxed-symbol.ts
arnog: I would avoid throwing an exception, although I would have expected that a constant would have had its type set when declared (the type being "inferred" based on the value at the time if necessary).

src/compute-engine/global-types.ts
```diff
@@ -541,15 +541,18 @@ export interface BoxedExpression {
...
    *
-   * For example, a number literal, a symbol with a constant value.
+   * Number literals, symbols with constant values, and numeric functions with constant
+   * subexpressions may all be considered *constant*, i.e.:
    * - `2` is constant
    * - `Pi` is constant
    * - `["Add", "Pi", 2]` is constant
-   * - `x` is not constant
-   * - `["Add", "x", 2]` is not constant
+   * - `x` is inconstant: unless declared with a constant value.
```
arnog, line 554: Perhaps "with a `constant` flag" would be more clear.

src/compute-engine/global-types.ts
```diff
@@ -1038,12 +1051,24 @@ export interface BoxedExpression {
    *
    * Infer the type of this expression.
    *
-   * If the type of this expression is already known, return `false`.
+   * Effective only for overall BoxedExpression types which are *non-constant* and therefore for
+   * which its value, and thereby type, can potentially vary.
+   *
+   * For symbols, inference may take place only for undeclared, or previously inferred symbols. For
```
arnog, line 1057: True, but with one caveat: a symbol declared without a type, or with a type `unknown` will still be subject to type inferrence.

src/compute-engine/global-types.ts
arnog: I believe that's correct

src/compute-engine/global-types.ts
```diff
@@ -1038,12 +1051,24 @@ export interface BoxedExpression {
...
+   * functions, only to those with an *inferred signature*.
    *
-   * If the type was not set, set it to the inferred type, return `true`
-   * If the type was previously inferred, widen it and return `true`.
+   * For a successful inference, *widens* the type for symbols (including creating a symbol
+   * definition if this is an _undeclared_ boxed-symbol), and for functions, narrows the *(return)
+   * type*. The return result will for this case be `true`.
+   *
+   * (Note that subsequent inferences can be made and will override previous ones if valid)
```
arnog, line 1064: Perhaps "refine" would be a better description than "override"

src/compute-engine/global-types.ts
```diff
@@ -1201,18 +1226,25 @@ export interface BoxedExpression {
    *
    * Equivalent to `expr.N().valueOf()`.
    *
+   * For functions, will only return non-undefined (i.e., compute the value) if the function is pure.
```
arnog, line 1229: I commented above about this, but I'm not sure why that would be the case. What would be more useful?

arnog: I'm also wondering if there isn't some confusion between a pure function and a pure function expression... That is `Sin` is a pure function, `["Sin", "Pi"]` is a pure function expression, `Random` is an impure function, `["Sin", ["Random"]]` is an impure function expression, `["Sin", "x"]` is a pure expression iff `"x"` is a constant.

test/compute-engine/numbers.test.ts
```diff
@@ -81,10 +81,10 @@ describe('PROPERTIES OF NUMBERS', () => {
       real: false
       rational: false
       integer: false
-      positive (>0): false
-      negative (<0): false
-      nonPositive (<=0): false
-      nonNegative (>=0): false
+      positive (>0): undefined
```
arnog, line 84: That seems incorrect. Shouldn't it be `false`? (same for the other sign properties)

test/compute-engine/numbers.test.ts
```diff
@@ -169,10 +169,10 @@ describe('PROPERTIES OF NUMBERS', () => {
...
```
arnog, line 172: Same here, I would expect `false`

test/compute-engine/numbers.test.ts
```diff
@@ -191,14 +191,14 @@ describe('PROPERTIES OF NUMBERS', () => {
...
```
arnog, line 194: I would expect `false` as well.

test/compute-engine/numbers.test.ts
```diff
@@ -213,10 +213,10 @@ describe('PROPERTIES OF NUMBERS', () => {
...
```
arnog, line 216: I would expect `false`

test/utils.ts
arnog: Looks good.

test/utils.ts
arnog: Looks good.

arnog (changes requested):
This is a great PR. Looks good overall. See comments, but the thing that bothers me the most is that some of the sign properties are for complex values are undefined, while they were `false` before. I'm not sure if that was an intentional change or a side effect.

The other thing, which is more of a heads up, is that the definition of `Sign` is a bit messy right now, as it includes overlapping concepts, such as the type, whether something is finite or not, etc... In general, I think it would be clearner and simpler if there were fewer "kinds" of sign. There is a pending PR that has those changes, I believe, and I'll have to check on its status and if it could be merged.

src/compute-engine/boxed-expression/boxed-number.ts
```diff
@@ -364,12 +380,21 @@ export class BoxedNumber extends _BoxedExpression {
 
     let s: number | undefined;
     if (typeof this._value === 'number') {
-      if (Number.isNaN(this._value)) return 'unsigned';
+      if (Number.isNaN(this._value)) return 'nan';
```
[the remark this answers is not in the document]
samueltlg: OK, good to know.

test/compute-engine/numbers.test.ts
samueltlg: Just realised that I forgot to publish the review yesterday; leaving only several isolated comments.

This change was intentional (if you see message for #238 8502b34) , on the basis that:
 - Generally,'undefined' (or *null*) is returned for boxed-expr. properties in which the property is inapplicable (e.g. `sgn` & other numeric properties for BoxedString; `symbol` for non-symbols; various properties, if the containing expr. is non-canonical...)
 - Therefore, since *sign*-ness does not apply to complex-numbers, at least in the traditional sense, then these properties are falling under category of 'inapplicability' and return undefined.

Indeed, a boolean could be returned in-place of 'undefined', but it appeared intuitive to me that this be returned here. Potentially, could save on some errors: if a complex number returned `isPositive: false`, and also jointly, `isNonPositive: false`, could see this being a bit confusing - particularly if the type of number was not checked prior. At least, 'undefined' circumvents cases such as that, and more-so follows prior rule.
Should be an easy reversion to 'undefined' though if you prefer the previous behaviour: will revert back!

samueltlg:
> This is a great PR. Looks good overall. See comments, but the thing that bothers me the most is that some of the sign properties are for complex values are undefined, while they were `false` before. I'm not sure if that was an intentional change or a side effect.
> 
> The other thing, which is more of a heads up, is that the definition of `Sign` is a bit messy right now, as it includes overlapping concepts, such as the type, whether something is finite or not, etc... In general, I think it would be clearner and simpler if there were fewer "kinds" of sign. There is a pending PR that has those changes, I believe, and I'll have to check on its status and if it could be merged.

Thanks . Have responded to initial comments; noticed that I didn't publish a review yesterday, meaning that they are bundled with a series of unpublished comments from then, too. Will also now add a small quantity of comments left to be added from yesterday.
Regardless of any upcoming amendments to the set of `sgn` changes, in general keep these changes as an interim fix? (fixes several, if you see #238 8502b342)

Think that, after resolving any current changes, & also pushing up the test-cases for `Number`, will add the changes for remaining forms such as 'Multiply','Divide', in a separate request; stop this one becoming bloated.

src/compute-engine/boxed-expression/canonical.ts
```diff
@@ -133,6 +134,11 @@ function numberForm(expr: BoxedExpression): BoxedExpression {
   //for explicitness in case things change)
   if (expr.isNumberLiteral) return expr.canonical;
 
+  // Ensure that all representations of the imaginary unit are represented with the BoxedNumber
+  // variant: this makes further simplifications more straightforward.
+  if (isImaginaryUnit(expr)) return expr.engine.I;
```
samueltlg, line 139: New op. added: similar below with `-i`.

arnog:
Regarding `holdUntil`...

> Only recently came to mind that the holdUntil symbol attribute is not accounted for during the check on operand values (which could potentially be symbols): should this be the case?

If `holdUntil` is never, the value of the symbol should be substituted during canonicalization.

See documentation for `holdUntil`:

```javascript
  /**
   * If the symbol has a value, it is held as indicated in the table below.
   * A green checkmark indicate that the symbol is substituted.

<div className="symbols-table">

| Operation     | `"never"` | `"evaluate"` | `"N"` |
| :---          | :-----:   | :----:      | :---:  |
| `canonical()` |    (X)    |              |       |
| `evaluate()`  |    (X)    |     (X)      |       |
| `"N()"`       |    (X)    |     (X)      |  (X)  |

</div>

  * Some examples:
  * - `ImaginaryUnit` has `holdUntil: 'never'`: it is substituted during canonicalization
  * - `x` has `holdUntil: 'evaluate'` (variables)
  * - `Pi` has `holdUntil: 'N'` (special numeric constant)
  * 
  * **Default:** `evaluate`
  */
  holdUntil: 'never' | 'evaluate' | 'N';
```

>Yet furthermore, if this is to be checked, i.e. that the 'holdUntil' value of a symbol-definition is never, then would it not be the case that symbols are substituted with values, before canonicalization, either partial or full, takes place?

It should be accounted for **during** canonicalization, as part of the canonicalization of the symbol.

> That being said, it would be the case that accounting for symbol values during application of canonical-forms, is unnecessary, since they would be substituted beforehand anyway (and any existing symbols would therefore have a 'holdUntil' value of 'evaluate', 'N', etc...)

Correct, no further check should be necessary during the canonicalization of arguments, since symbols should have be substituted by then if indicated.

src/compute-engine/boxed-expression/boxed-symbol.ts
arnog: Yes, I think so. Might still be useful as an assert (or running the test suite with the assert to validate that my assumption is correct... :) )

test/compute-engine/numbers.test.ts
arnog: OK, you've convinced me. Initially, I was thinking of emulating the behavior of `NaN`, where both `NaN > 0` and `NaN < 0` return false, but obviously with this syntax returning `undefined` is not an option.

So let's go with that. The only think to be careful of is that some sign calculations in `arithmetic.ts` could become invalid.

test/compute-engine/__snapshots__/arithmetic.test.ts.snap
```diff
@@ -335,6 +335,7 @@ N-mach    = [3.00416602394643,7.3890560989306495,54.59815003314423]
 
 exports[`EXP Exp -1 1`] = `
 box       = ["Exp", -1]
```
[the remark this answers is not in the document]
arnog: Do you mean that the canonical form of `["Exp", -1]` is now `["Divide", 1, "ExponentialE"]`?

test/compute-engine/ascii-math.test.ts
```diff
@@ -120,9 +120,9 @@ describe('POWERS/ROOTS', () => {
     expect(check('x+(-1)^2')).toMatchInlineSnapshot(`x + (-1)^2`);
   });
   it('should serialize other powers', () => {
-    expect(check('x^{0}')).toMatchInlineSnapshot(`1`);
```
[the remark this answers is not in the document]
arnog: Yes, I think that `x^0` should stay as is after canonicalization, even if `x` is constant. Only if `x` was `holdUntil = "never"` should it be substituted.

test/compute-engine/ascii-math.test.ts
```diff
@@ -258,7 +258,7 @@ describe('PRECEDENCE', () => {
     expect(check('1+2^3')).toMatchInlineSnapshot(`1 + 2^3`);
     expect(check('(1+2)^3')).toMatchInlineSnapshot(`(1 + 2)^3`);
     expect(check('1^2+3')).toMatchInlineSnapshot(`1 + 3`);
-    expect(check('1^{2+3}')).toMatchInlineSnapshot(`1`);
```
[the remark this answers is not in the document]
arnog: Looks good.

src/compute-engine/boxed-expression/sgn.ts
```diff
@@ -44,63 +44,110 @@ export function infinitySgn(s: Sign | undefined): boolean | undefined {
   );
 }
 
-// > 0
+/**
+ * Sign `s` is > 0.
+ *
+ * :::info[Note]
+ * Returns `undefined` for cases where the given sign is either non-applicable to real numbers
```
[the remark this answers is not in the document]
arnog: Yes, sounds good, as per above.

src/compute-engine/boxed-expression/boxed-symbol.ts
```diff
@@ -68,27 +69,47 @@ import { BoxedType } from '../../common/type/boxed-type';
...
+ * 'isNaN', 'isInfinity' 'isPositive' & 'isNegative'.
+ *   Notably, this is the case for properties which relate to a *range* of values or 'assumptions'
+ *   about the symbol (e.g. 'is..Positive/Negative'); or in which looking up the value directly is
+ *   not necessary to compute the result (e.g. 'isFinite').
+ *
+ *
+ * Methods which require consultation to a definition value also trigger binding: comparison methods
+ * ('is', 'isSame', 'isEqual', 'isLess'...), 'evaluate' & 'N', and methods pertinent to
+ * 'collection'-valued symbols such as 'contains' & 'subsetOf'.
+ *
```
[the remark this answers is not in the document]
arnog: Yes, that's correct.

src/compute-engine/boxed-expression/abstract-boxed-expression.ts
```diff
@@ -224,7 +224,7 @@ export abstract class _BoxedExpression implements BoxedExpression {
   toLatex(options?: Partial<SerializeLatexOptions>): LatexString {
     // We want to use toMathJson(), not .json, so that we have all
     // the digits for numbers, repeated decimals
-    const json = this.toMathJson();
+    const json = this.toMathJson({ prettify: options?.prettify ?? true });
```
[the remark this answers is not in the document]
arnog: Looks good.

src/compute-engine/boxed-expression/box.ts
```diff
@@ -191,17 +191,15 @@ export function boxFunction(
...
-        options.canonical.includes('Number')));
-
+  const canonicalNumber = structural === false && options.canonical === true;
+
+  // If canonical, handle cases of various expression structures being able to be cast as
+  // BoxedNumbers (some cases of Negate, Rational, Divide, Complex), or 'de-number' some 'borderline
+  // invalid' boxed number-like expressions
+  // (!@note: this procedure is similarly repeated within the 'number' CanonicalForm, but the
+  // numberForm variant more simply applies to fully BoxedExprs., and during partial canonicalization
+  // only)
```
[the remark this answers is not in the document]
arnog: That's correct, good catch. This check is not necessary.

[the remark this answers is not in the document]
arnog: Correct.

src/compute-engine/boxed-expression/boxed-function.ts
```diff
@@ -197,10 +197,8 @@ export class BoxedFunction extends _BoxedExpression {
 
   get isPure(): boolean {
     if (this._isPure !== undefined) return this._isPure;
-    if (!this.isCanonical) {
-      this._isPure = false;
-      return false;
-    }
+    if (this.isCanonical === false) return this.canonical.isPure;
```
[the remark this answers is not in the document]
arnog: I'm not 100% sure, but that's reasonable.

src/compute-engine/boxed-expression/arithmetic-power.ts
```diff
@@ -31,38 +32,182 @@ export function asRadical(expr: BoxedExpression): Rational | null {
...
+ * Both the given base and exponent can either be canonical or non-canonical: with fully
+ * canonicalized args. lending to more simplifications.
+ * 
+ * Returns a canonical expr. is both operands are canonical.
+ 
+ * @export
+ * @param a
+ * @param b
+ * @returns
+ */
```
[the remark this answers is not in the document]
arnog: > More consistently, checks for _constant_ symbol values for operations, such as whether 'b' is '1' in 'a^b'

The value of symbols should not be considered during canonicalization, even constant ones. That's because their value (and even the fact they are constant) depends on binding. Ideally, canonicalization would be independent of binding, particularly of binding of symbols. Note that currently that is not the case, but that is a fundamental error that I've only fairly recently realized. Eventually, the canonicalization should happen by applying a series of rules to expressions, much like simplification rules, and these should not depend on binding having occurred.

> -**warning**: this may be *incorrect* behaviour: since the 'holdUntil' attribute is not looked to (and this may be desirable). Furthermore, if this is an oversight, then it may be the case that symbols with a 'never' holdUntil value will be substituted with their values before canonicalization/canonical-forms: therefore accounting for symbol values here, or within any other canonical-form, may be unnecessary & inappropriate.

As per discussion previously, I don't think you need to look at the `holdUntil` attribute. Symbols with a `holdUntil = "never"` will get substituted during their canonicalization and that is correct.

> !Function 'pow' - the first half more or less - is a replication of what now takes place within canonicalPower: but is less thorough, and  may contain a few inaccuracies. Likely, would benefit from replacing its first half with a call to 'canonicalPower'; which it anyway  replicates.

yes, that sounds reasonable.

src/compute-engine/boxed-expression/arithmetic-power.ts
```diff
@@ -31,38 +32,182 @@ export function asRadical(expr: BoxedExpression): Rational | null {
...
+  if (b.is(0.5)) return canonicalRoot(a, 2);
   const r = asRational(b);
-  if (r !== undefined && r[0] === 1) return canonicalRoot(a, ce.number(r[1]));
 
-  return ce._fn('Power', [a, b]);
+  //1/3, 1/4...
+  if (r !== undefined && r[0] === 1 && r[1] !== 1)
+    return canonicalRoot(a, ce.number(r[1]));
+
+  return unchanged();
```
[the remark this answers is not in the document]
arnog: Meh... I guess...

src/compute-engine/boxed-expression/arithmetic-power.ts
```diff
@@ -31,38 +32,182 @@ export function asRadical(expr: BoxedExpression): Rational | null {
...
+ * canonicalized args. lending to more simplifications.
+ * 
+ * Returns a canonical expr. is both operands are canonical.
+ 
+ * @export
+ * @param a
+ * @param b
+ * @returns
+ */
 export function canonicalPower(
```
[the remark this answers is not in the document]
arnog: Calling `canonicalPower()` in `pow()` sounds like a good idea. This is not redundant, because `pow(a, b)` can be called with `a` and `b` which may not be canonical (i.e. it's used internally to make some intermediate calculations).

src/compute-engine/boxed-expression/boxed-symbol.ts
samueltlg: > Yes, I think so. Might still be useful as an assert (or running the test suite with the assert to validate that my assumption is correct... :) )

Oh; no, this condition was added as a fix so was apparently necessary (but of course was none-the-wiser to fact that 'inferredType' should jointly be set to *false*).
I think that, maybe  what's going on is that the *type* is being set, but the *isInferred* flag is not being adjusted?

You can observe why this fix (36edf95) was added with the following example:
```typescript
ce.declare('x', {constant:true,value:2});
ce.parse('x + 2'); // -> 'Error: The type of the constant "x" cannot be changed' (takes place at `BoxedSymbol.infer()` as above; this being called from 'checkNumericArgs', as called from 'makeNumericFunction'
```
So seems to surface during the full canonicalization of 'Add' (& likely other fundamental/numeric functions)

This 'fix' stops the error for now, instead leading to 'symbol.infer()' returning 'false'

At the close of this pull-request, I could either leave this as a temporary fix; or if you imagine it's a fairly simple fix, you could point me in the right direction (something like `BoxedSymbolDefinition` constructor, I imagine), at which *inferredType* is to be set to `false` (for declaration/definition of a constant symbol), & I'll add a patch commit?

src/compute-engine/global-types.ts
samueltlg: OK 👍

test/compute-engine/numbers.test.ts
samueltlg: Ok, cool.
Will make a note on sign-calculations potentially breaking, for something to look at a future date, in-case some related errs./unexpected behaviour pop-up 👍

test/compute-engine/__snapshots__/arithmetic.test.ts.snap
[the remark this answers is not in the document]
samueltlg: > Do you mean that the canonical form of `["Exp", -1]` is now `["Divide", 1, "ExponentialE"]`?

Yes

src/compute-engine/boxed-expression/canonical.ts
```diff
@@ -105,13 +106,93 @@ function invisibleOperatorForm(expr: BoxedExpression) {
...
+    const { numericValue } = op1;
+    if (numericValue !== null)
+      return ce.number(
+        typeof numericValue === 'number' ? -numericValue : numericValue.neg()
+      );
+  }
+
+  return ops.every((op, index) => op === expr.ops![index])
+    ? expr
+    : ce._fn(name, ops, { canonical: false });
```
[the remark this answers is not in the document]
samueltlg: That's good.
Will for now just stick with the added `i`/`-i` boxing: but maybe the 'Add'/'Multiply' (i.e. for Complex) something to consider later

src/compute-engine/boxed-expression/arithmetic-power.ts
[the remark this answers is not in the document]
samueltlg: > The value of symbols should not be considered during canonicalization, even constant ones

Cleared-up now.
It was something that didn't come to mind until recently, namely that due to the 'holdUntil' attribute, symbols would be substituted with values prior to/at canonicalization (->& therefore no need to check).

Out of curiosity though, did think that it was OK to check whether a symbol is a constant (`isConstant`) without it triggering binding? If I recall: think its associated documentation states this to be the case...?

src/compute-engine/boxed-expression/boxed-function.ts
```diff
@@ -210,14 +208,24 @@ export class BoxedFunction extends _BoxedExpression {
...
+  get constantValue(): number | boolean | string | object | undefined {
+    return this.isConstant ? this.value : undefined;
+  }
+
+  /**
+   *
+   * @inheritdoc
+   */
+  get value(): number | boolean | string | object | undefined {
+    if (!this.isPure) return undefined;
```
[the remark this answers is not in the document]
arnog: ok, let's go with modifying `.value` to return undefined if the expression is not a boxed number or a boxed symbol. This should not require other internal changes.

src/compute-engine/boxed-expression/arithmetic-power.ts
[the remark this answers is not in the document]
samueltlg: I see. In it goes.

samueltlg:
> Regarding `holdUntil`...
> 
> > Only recently came to mind that the holdUntil symbol attribute is not accounted for during the check on operand values (which could potentially be symbols): should this be the case?
> 
> If `holdUntil` is never, the value of the symbol should be substituted during canonicalization.
> 

Just thought... does this substitution take place for partial canonicalization/CanonicalForms?

samueltlg:
Have a closing list of changes to make now; along with the final Number-form tests to be added up.
Resolution points remaining points as just commented:
- `BoxedExpr.value`
- Constant-symbol, *inferredType* 'bug'
- `["Exp", -1] is now ["Divide", 1, "ExponentialE"]` is acceptable?

After clearing those up & making final changes + tests, should be good to go.

src/compute-engine/boxed-expression/boxed-symbol.ts
arnog: Ah, OK, yeah that was incorrect. I fixed it by changing line 148 in `boxed-symbol.ts` to `this.inferredType = !this.isConstant`

arnog:
> BoxedExpr.value
Let's have `BoxedExpr.value` return `undefined` if the expression is not a literal or a symbol with a value.

> Constant-symbol, inferredType 'bug'
I've checked in a fix.

> ["Exp", -1] is now ["Divide", 1, "ExponentialE"] is acceptable?
Yes

samueltlg:
The 'value' property, and other changes will be present for currently 😄. 
May also be worth re-updating the doc. after next fix-up; some still awaits req.-review/update

PR #241, opened 2025-04-15
Fix/test: amendments to #238 (CanonicalForms PR), 'Number' form tests, some arch.

samueltlg:
Only a smaller one this time

But a few key points/queries which will be highlighted in comments:

- Not sure if, as part of the planned re-arch. of canonicalization, that these `CanonicalForm` *functions* will remain (i.e. as function-based replacement rules), or be replaced by individual, string-based rules. Of note though, is that the thus-far published forms should be useful guidance in achieving this, & should reduce the workload substantially.
  (In some cases though, I have pondered how/if string-based replacement rules could achieve this (so easily): considering the substantial set of conditions required upon some operations.
  I.e. you could imagine: `x_{type:number;!isFunctionExpression;...}^0  ->  1`, or `a_{AnyInfinity}^b_{type:complex; real > 0}  ->  ComplexInfinity`, to be verbose...)

- Have wondered, with regards to the discussion https://github.com/cortex-js/compute-engine/pull/238#discussion_r2034335185, particularly `The value of symbols should not be considered during canonicalization, even constant ones...`:
  - Is it the case that this should be amended slightly to `The value of *non-canonical/bound* symbols...`. Because, as will be remarked in the upcoming review, `Pi`, for example, being a library-defined ID., appears to typically be 'always' bound/canonical: such that, unlike the case for user-defined/non-default library constants (which as stated should _not_ be bound at canonicalization), it seems reasonable that `Pi^0`, `-Pi^Infinity`, etc. *should* apply?

- Believe that there is a current bug for SymbolDefinitions.
  It appears that, `holdUntil` being set to `never` should only be permitted when the *constant* switch is also `true`.
  Otherwise, these non-constant symbols get substituted at canonicalization, resulting in inconsistent/variable full canonical-form?

It should be 1-2 weeks until I can finish-up the final/remaining forms PR (at least to the same degree of attention as here), but that will come.

In the next post, I will leave some open questions that have popped-up throughout my completing this request & again having stumbled into many things. (Most of which related to what's going on here).

### DIFF
--- a/src/compute-engine/boxed-expression/arithmetic-power.ts
+++ b/src/compute-engine/boxed-expression/arithmetic-power.ts
@@ -68,17 +68,13 @@ export function canonicalPower(
   // here)
   if (
     b.isFunctionExpression ||
-    !b.isConstant || //(!notably, should exclude symbol assumptions)
+    b.symbol !== null ||
     !b.type.matches('number' as Type)
   )
     return unchanged();
 
-  // 'a' is constant and of type 'number'. Used often.
-  // (note that this could still be a numeric expression, too...)
-  const aIsNum = a.isConstant && a.type.matches('number' as NumericType);
-
   // Zero as base
-  if (aIsNum && a.is(0)) {
+  if (a.isNumberLiteral && a.is(0)) {
     if (b.type.matches('imaginary' as NumericType) || b.isNaN) return ce.NaN;
 
     if (b.is(0)) return ce.NaN;
@@ -97,9 +93,19 @@ export function canonicalPower(
     return unchanged(); // No other canonicalization cases with this base
   }
 
+  // 'a'/base has an associated number value (excludes numeric functions)
+  // (this should at this stage include library-defined symbols such as 'Pi')
+  //@note: include 'Negate', because this could be wrapped around a number-valued symbol, such as
+  //'Pi'...
+  //^there could exist other exceptions: perhaps consider a util. such as 'maybeNumber'?
+  const aIsNum =
+    a.type.matches('number' as NumericType) &&
+    (a.isFunctionExpression === false || a.operator === 'Negate');
+
   // Zero as exponent
   if (b.is(0)) {
-    if (aIsNum) return a.isFinite ? ce.One : ce.NaN;
+    // If 'isFinite' is a boolean, then 'a' has a value.
+    if (aIsNum && a.isFinite !== undefined) return a.isFinite ? ce.One : ce.NaN;
     return unchanged();
   }
 
@@ -109,7 +115,8 @@ export function canonicalPower(
   if (aIsNum && a.is(1)) return b.isFinite ? ce.One : ce.NaN;
 
   // One as exponent
-  if (b.is(1)) return a;
+  // (Permit the base to be a FN-expr. here, too...)
+  if (b.is(1) && a.type.matches('number' as NumericType)) return a;
 
   // -1 exponent
   if (b.is(-1)) {
@@ -176,8 +183,8 @@ export function canonicalPower(
     return ce.NaN;
   }
 
-  //'Infinity^Complex'
-  if (a.isInfinity) {
+  //'AnyInfinity^{~oo}' (i.e. ComplexInfinity)
+  if (a.isNumberLiteral && a.isInfinity) {
     // If the exponent is pure imaginary, the result is NaN
     //(↓fix?:ensure both these cases narrow down to 'b' being a num./symbol literal)
     if (b.type.matches('imaginary' as NumericType)) return ce.NaN;
--- a/src/compute-engine/boxed-expression/arithmetic-power.ts
+++ b/src/compute-engine/boxed-expression/arithmetic-power.ts
@@ -61,8 +61,7 @@ export function canonicalPower(
     ce._fn('Power', [a, b], { canonical: fullyCanonical });
 
   // Exclude cases - which may otherwise be valid - of the exponent either: being a function (e.g.
-  // '0 + 0'), a non-constant value (symbols with 'assumed' values, impure functions), or a
-  // non-numeric type (no concern for this).
+  // '0 + 0'), a symbol, or non-numeric expr. type (e.g. string).
   //
   // @consider:possible exceptions where function-expressions are reasonable :Rational,Half,
   // Negate... (However, provided that canonicalNumber provided prior, should not be missing anything
@@ -273,93 +272,12 @@ export function pow(
 
   if (typeof exp !== 'number') exp = exp.canonical;
 
+  // 'canonicalPower' deals with a set of basic operations.
+  // If the result is not 'Power', can assume an op. has taken place
+  const canonicalResult = canonicalPower(x, ce.box(exp));
+  if (canonicalResult.operator !== 'Power') return canonicalResult;
+
   const e = typeof exp === 'number' ? exp : exp.im === 0 ? exp.re : undefined;
-
-  // x^0 = 1
-  if (e === 0) return ce.One;
-  // x^1 = x
-  if (e === 1) return x;
-
-  if (e === -1) {
-    // (-∞)^-1 = 0
-    if (x.isInfinity && x.isNegative) return ce.Zero;
-
-    // (-1)^-1 = -1
-    if (x.is(-1)) return ce.NegativeOne;
-
-    // 0^-1 = ~∞
-    // This is not strictly true, as 0^-1 may be undefined, but is convenient in some contexts where the base is assumed to be positive.
-    if (x.is(0)) return ce.ComplexInfinity;
-
-    // 1^-1 = 1
-    if (x.is(1)) return ce.One;
-
-    // ∞^-1 = 0
-    if (x.isInfinity && x.isPositive) return ce.Zero;
-
-    return x.inv();
-  }
-
-  if (e === Number.POSITIVE_INFINITY) {
-    // 0^∞ = 0
-    // Because for all complex numbers z near 0, z^∞ -> 0.
-    if (x.is(0)) return ce.Zero;
-
-    // 1^∞ = NaN
-    // Because there are various cases where lim(x(t),t)=1, lim(y(t),t)=∞ (or -∞), but lim( x(t)^y(t), t) != 1.
-    if (x.is(1)) return ce.NaN;
-
-    // (-1)^∞ = NaN
-    // Because of oscillations in the limit.
-    if (x.is(-1)) return ce.NaN;
-
-    if (x.isInfinity) {
-      if (x.isPositive) return ce.PositiveInfinity;
-      if (x.isNegative) return ce.NaN;
-    }
-  }
-
-  if (e === Number.NEGATIVE_INFINITY) {
-    if (x.is(-1)) return ce.NaN;
-    if (x.isInfinity) {
-      if (x.isPositive) return ce.Zero;
-      if (x.isNegative) return ce.NegativeInfinity;
-    }
-  }
-
-  if (typeof exp !== 'number') {
-    if (exp.isInfinity && !exp.isPositive && !exp.isNegative) {
-      // b^~∞ = NaN
-      // Because b^z has no limit as z -> ~∞.
-      return ce.NaN;
-    }
-
-    if (x.isInfinity) {
-      // If the exponent is pure imaginary, the result is NaN
-      if (exp.type.matches('imaginary')) return ce.NaN;
-      if (exp.type.matches('complex') && !isNaN(exp.re)) {
-        if (exp.re > 0) return ce.ComplexInfinity;
-        if (exp.re < 0) return ce.Zero;
-      }
-    }
-  }
-
-  //   // if (this.isNegative) {
-  //   //   if (exp % 2 === 1) return this.neg().pow(exp).neg();
-  //   //   if (exp % 2 === 0) return this.neg().pow(exp);
-  //   // }
-
-  if (e === Number.POSITIVE_INFINITY) {
-    if (x.isGreater(1)) return ce.PositiveInfinity;
-    if (x.isPositive && x.isLess(1)) return ce.Zero;
-  }
-  if (e === Number.NEGATIVE_INFINITY) {
-    if (x.isGreater(1)) return ce.Zero;
-    if (x.isPositive && x.isLess(1)) return ce.PositiveInfinity;
-  }
-
-  if (typeof exp !== 'number' && exp.operator === 'Negate')
-    return pow(x, exp.op1, { numericApproximation }).inv();
 
   // @todo: this should be canonicalized to a number, so it should never happen here
   if (x.symbol === 'ComplexInfinity') return ce.NaN;
--- a/src/compute-engine/boxed-expression/arithmetic-power.ts
+++ b/src/compute-engine/boxed-expression/arithmetic-power.ts
@@ -60,8 +60,19 @@ export function canonicalPower(
   const unchanged = () =>
     ce._fn('Power', [a, b], { canonical: fullyCanonical });
 
-  // Exclude cases - which may otherwise be valid - of the exponent either: being a function (e.g.
-  // '0 + 0'), a symbol, or non-numeric expr. type (e.g. string).
+  if (a.operator === 'Power') {
+    const [base, aPow] = a.ops!;
+    return ce._fn('Power', [
+      base,
+      ce.box(['Multiply', aPow, b], {
+        canonical: fullyCanonical || 'Power',
+      }),
+    ]);
+  }
+
+  // Onwards, the focus on operations is where is a *numeric* exponent.
+  // Therefore, exclude cases - which may otherwise be valid - of the exponent either: being a function (e.g.
+  // '0 + 0'), a symbol, or of a non-numeric type.
   //
   // @consider:possible exceptions where function-expressions are reasonable :Rational,Half,
   // Negate... (However, provided that canonicalNumber provided prior, should not be missing anything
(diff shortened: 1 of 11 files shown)


src/compute-engine/boxed-expression/boxed-function.ts
```diff
@@ -681,7 +681,7 @@ export class BoxedFunction extends _BoxedExpression {
     }
 
     // root(sqrt(a), c) -> root(a, 2*c)
-    if (this.operator === 'Sqrt' || this.operator === 'Root') {
+    if (this.operator === 'Sqrt') {
```
samueltlg, line 684: (^WRT commit message)
I did remove `!BoxedExpr.value now altogether returns 'undefined' for non-literal
  boxed-expr. types (e.g. functions)`, (an older edit) from the message, but this has somehow persisted.

src/compute-engine/boxed-expression/canonical.ts
```diff
@@ -22,6 +22,17 @@ export function canonicalForm(
...
+  // Automatically, this involves the substitution of the symbol with its value, if it is a
+  // constant-flagged symbol, with a 'holdUntil' attribute of 'never'
+  // (@note: the reasoning for carrying this out here is because in some sense, firstly,
+  // 'CanonicalForm' can be regarded as producing a 'canonical' expression, (albeit a 'custom' one;
+  // notably also with the resulting expr. having an 'isCanonical' value of 'true'). Secondly &
+  // following from this, symbol canonicalization (and substitution where appropriate) facilitates
+  // several simplifications which would otherwise not be made: for example 'x^y' where 'y=0', for
+  // canonicalPower.
+  expr = symbolForm(expr);
+
```
samueltlg, line 35: I have seen that a `Symbol` CanonicalForm has been thought about/mentioned via comments in one or two places, but it seems to me that doing this consistently (for partial canonicalization) makes sense, given the reasoning in this associated comment, and the one prior. What do you think ?

src/compute-engine/boxed-expression/arithmetic-power.ts
```diff
@@ -97,9 +93,19 @@ export function canonicalPower(
...
+  //^there could exist other exceptions: perhaps consider a util. such as 'maybeNumber'?
+  const aIsNum =
+    a.type.matches('number' as NumericType) &&
+    (a.isFunctionExpression === false || a.operator === 'Negate');
+
   // Zero as exponent
   if (b.is(0)) {
-    if (aIsNum) return a.isFinite ? ce.One : ce.NaN;
+    // If 'isFinite' is a boolean, then 'a' has a value.
+    if (aIsNum && a.isFinite !== undefined) return a.isFinite ? ce.One : ce.NaN;
```
samueltlg, line 108: Here, `isFinite` being 'undefined' is also acting as a check as to whether, if a symbol, it has a bound (numeric) value.
Is there a method/accessor equivalent to `hasValue()` - this has come to have use at least in a couple of scenarios for myself.
I know that the `isConstant` symbol method triggers binding (although the outdated in-lined doc. states otherwise); I guess then that it is not possible ?

(Do see PR conversation post 'open questions' regarding this point (although, may not be posted at time you read this))

test/compute-engine/canonical-form.test.ts
```diff
@@ -286,11 +321,18 @@ describe('CANONICAL FORMS', () => {
...
       expect(checkPower('16^x')).toMatchInlineSnapshot(`
         box        = ["Power", 16, "x"]
         canonForms = 1
         canonical  = 1
       `);
+      //!@note: this should be ok to apply, despite 'Pi' being a constant which does *not* have a
+      //!'holdUntil' value set to 'never', because... this is a library/engine-level defined
+      //!constant in which its type/value is always known: even in non-canonical expressions.
+      //I.e., if this were a user-defined constant, this would not be OK, since its value would have
+      //to be referenced.
```
samueltlg, line 335: This case relates to the second bullet printed in the PR opening.
Seems reasonable to me that this proceeds, as an exception, given that it is an engine/inbuilt *library* definition?

test/compute-engine/canonical-form.test.ts
```diff
@@ -342,15 +393,41 @@ describe('CANONICAL FORMS', () => {
...
+      `);
+
+      //↓🙁: cannot be simplified, at least via the route of checking the 'type' of the base/'-\pi',
+      //because is a non-canonical function-expr. (non-bound to 'Power' definition)
+
+      // expect(checkPower('{-\\pi}^1')).toMatchInlineSnapshot(`
+      //   box        = ["Power", "Pi", 1]
+      //   canonForms = Pi
+      //   canonical  = Pi
+      // `);
```
samueltlg, line 418: If being strict & proper, sadly this cannot apply, despite `Pi` being a library-definition constant & having a known value/type, because the containing `Negate` expression is non-canonical, & therefore unable to reveal its type. 
(Also see PR conversation 'open questions' post)

test/compute-engine/canonical-form.test.ts
```diff
@@ -342,15 +393,41 @@ describe('CANONICAL FORMS', () => {
...
+       * Control
+       */
+      //!@note:
+      //Not currently simplified, despite 'j' being an 'integer'-typed constant, because the base is
+      //a non-canonical function and therefore unbound to a definition; consequently with its 'type'
+      //non-determinable.
+      expect(checkPower('{-j/4}^1')).toMatchInlineSnapshot(`
+        box        = ["Power", ["Divide", ["Negate", "j"], 4], 1]
+        canonForms = ["Power", ["Divide", ["Negate", "j"], 4], 1]
+        canonical  = ["Multiply", ["Rational", -1, 4], "j"]
```
samueltlg, line 430: A similar case to above

samueltlg:
(Have to go; but there's a few open-ended Q's/queries to post here, pending, in relation to changes: particularly relation between canonicalization/binding, and (function) types)

samueltlg:
Some open questions:

Regarding this points made in https://github.com/cortex-js/compute-engine/pull/238#discussion_r2034335185

- '`Ideally, canonicalization would be independent of binding, particularly of binding of symbols`'
  - The potential issue of this being the case is... ? (I'm guessing, scopes? i.e., by the time it comes to working with the expression, or containing expression further (such as evaluation), this may occur in the context of a new scope: meaning that a new definition maybe should be associated with the symbol instance? (too late, if it is already bound?
  - If these two were to be separated, then what would canonicalization of symbols entail, in lieu ?
  - Generally, for the purposes of a workaround for avoiding binding & obtaining a symbol's its definition/its info. (in the current scope), is requesting the `canonical` variant of symbol, acceptable?
     This in turn, begs the question, is a/the problem with binding in this context, the binding of the symbol *instance*, or, what it entails: namely the creation of an associated _definition_ (where this would best be/ideally inferred) ?

   Also just in case worth asking; is it at all possible to obtain the symbol's *type* without binding? Or is it all the same (the 
   definition being bound required for determining any of *isConstant*, *type*, *holdUntil*, et cetera) ?

- As remarked/hinted at in some code comments already:
  - Some partial-canonicalization/CanonicalForm operations cannot reliably take place because *non-canonical* function-expressions cannot have their `type` ascertained (seemingly due to not being definition-bound): notably even this is clear on inspection, for instance:
     ```typescript
      ce.parse('{-\pi}^Infinity', {canonical: 'Power'}); // Remains unchanged
      ce.parse('{\pi + 1}^1', {canonical: 'Power'});     // Remains the same
     ```
For both of these cases, the revised `canonicalPower` is careful not to simplify these two, because the base operand cannot be verified in each case to be *numeric* (have a `number` or more refined) type.
It did come to mind then, at least two workarounds:
  - Requesting/peeking the canonical variant (if the base is a *function-expression*).
     Clearly, this non-ideal (if not a tad problematic), and further, certainly certainly non-optimal.
  - Perhaps better, if feasible, would be:
    - In a similar way to how 'common' numeric operations/fundamental arithmetic ones have a special handling with reference to canonicalization (the '*makeNumericFunction*' function, for instance), would it not be reasonable for these common operations to have a shortcut for peeking their type (based on arguments), i.e. without first being definitely bound to a definition? Imagine that this would have reasonable usefulness: not to mention these outlined cases.

There are a couple more points, but enough for now... !

src/compute-engine/boxed-expression/boxed-function.ts
arnog: OK, not sure what happened...

src/compute-engine/boxed-expression/canonical.ts
arnog: Yes, that's reasonable.

src/compute-engine/boxed-expression/arithmetic-power.ts
arnog: Isn't this the same as checking that `expr.value !== undefined`?

test/compute-engine/canonical-form.test.ts
arnog: I agree.

test/compute-engine/canonical-form.test.ts
arnog: Yes, this might need to be addressed by introducing the concept of **global scopes**...

test/compute-engine/canonical-form.test.ts
arnog: Yes, this might need to be addressed by introducing the concept of **global scopes**...

arnog:
OK, I've given some more thoughts on canonicalization and binding.

I think that binding and canonicalization (including signature validation) can all happen at the same time, namely during boxing/parsing. 

However, what will need to change is how values are associated with identifiers. They are currently associated with the definition record of the identifier, but this will need to move to a separate "frame" record in order to support recursion. 

The value of constants will still be stored in their definition records, though. This means that the value of constants (and other properties, such as sign) can be considered during canonicalization (but only for constants and literals). The type of symbols (constant or not) is also available during canonicalization, as well as other symbol attributes, such as `holdUntil`.

Accessing the value or other properties depending on the value of non-constant identifiers during canonicalization should return undefined, or possibly throw an error (this can be achieved by disabling the frame stack at the start of canonicalization).

samueltlg:
> OK, I've given some more thoughts on canonicalization and binding.
> 
> I think that binding and canonicalization (including signature validation) can all happen at the same time, namely during boxing/parsing.
> 
> However, what will need to change is how values are associated with identifiers. They are currently associated with the definition record of the identifier, but this will need to move to a separate "frame" record in order to support recursion.
> 
> The value of constants will still be stored in their definition records, though. This means that the value of constants (and other properties, such as sign) can be considered during canonicalization (but only for constants and literals). The type of symbols (constant or not) is also available during canonicalization, as well as other symbol attributes, such as `holdUntil`.
> 
> Accessing the value or other properties depending on the value of non-constant identifiers during canonicalization should return undefined, or possibly throw an error (this can be achieved by disabling the frame stack at the start of canonicalization).

That would be good, most likely, if signature validation could remain (this stage).

I think I get the gist of what you are saying there... the 'frames' seemingly being like a scope stack, but rooted at the original definition (explicitly declared or otherwise)? And this would be instead of, or in addition, to a 'global symbol table/scope'?
And is pertinent to functions (i.e. the binding), or function-definition bound symbols, too?
Would this help in being able to determine the type of numeric-functions during canonicalization, as in the prior given `ce.parse('{-\pi}^Infinity', {canonical: 'Power'})` example ?

Does sound useful that that extended properties will be available for non-constant & symbols for canonicalization :thumbsup:

src/compute-engine/boxed-expression/arithmetic-power.ts
samueltlg: > Isn't this the same as checking that `expr.value !== undefined`?

Would think... but not quite, since `value` triggers binding, currently. And so 'isFinite' is a funny check here for determining whether a value is present (with it not triggering binding).
Hence why maybe a `hasValue` or equivalent would be useful as things stand: but maybe not depending on incoming changes...

arnog:
> the 'frames' seemingly being like a scope stack, but rooted at the original definition (explicitly declared or otherwise)?

The frames are an evaluation context. They are arranged in a stack. A new frame is pushed on the stack when a function is evaluated (this could be optimized to be only when a function with named arguments or local variables is executed). The frame only keep track of the current values of named arguments and local variables.

They are distinct from the scope, which is a lexical context. A new scope is pushed on the scope stack when a new declaration context is boxed (canonicalized), not when it's evaluated. The scope keeps track of the type and attributes of identifiers (constant, holdUntil, etc...)

> And this would be instead of, or in addition, to a 'global symbol table/scope'?

This would be instead of. There might still be some value in a global table/scope, but not clear what at this point.

> And is pertinent to functions (i.e. the binding), or function-definition bound symbols, too?

Frames are not used for binding, only scopes are.

> Would this help in being able to determine the type of numeric-functions during canonicalization, as in the prior given `ce.parse('{-\pi}^Infinity', {canonical: 'Power'})` example ?

Yes, because the type information and attributes (constant) (for `Pi`) would be available during canonicalization. In theory, even the value of constants could be used during canonicalization, but attempting to use the value of non-constants would probably throw an error.

samueltlg:
OK, think I've got it. The the evaluation contexts are still *informed* by/connected to scopes, no? In order that values of local variables/symbols can be looked-up/determined?
The issue with the `'{-\pi}^Infinity'` case at canonicalization is not `Pi`'s attributes in this case, but rather that of 'Negate', i.e. because The Negate expression `-\pi` is not canonical & therefore not bound, it's *type* cannot be calculated as 'Number', for example.

src/compute-engine/boxed-expression/arithmetic-power.ts
arnog: OK, thanks for bringing this up. This lead me to revisit the various properties of `BoxedExpression` and which ones require a canonical form, or will automatically cause binding. It seems there are discrepancies between the documentation and what the code does, and perhaps the rules need to be revisited for improved consistency.

How about this:
- operations that are purely structural can be done on unbound (non-canonical) expressions (e.g. `expr.subs()`, `expr.toString()`, `expr.isSame()`, etc...)
- boxed expressions are never bound as a side effect, they are only bound when their canonical form is created (i.e. at construction)
- operations/properties that rely on the value of the expression return undefined if the expression is not canonical (not bound), or a function expression. Otherwise, for number literals and symbols with an associated value, the value is used.

arnog:
> The the evaluation contexts are however still informed by/connected to scopes, no? In order that values of local variables/symbols can be looked-up/determined? Think that is where I was getting at.

The frames ("evaluation context") keep track of the values of local variables and arguments (non constant identifiers). They do not need the scopes (lexical context) to do so.

Each symbol or function expression *is* bound to a record in a scope, which can be used to access metadata about it (is it a constant, what is its type, etc...).

> The issue with the '{-\pi}^Infinity' case at canonicalization is not Pi's attributes in this case, but rather that of 'Negate', i.e. because The Negate expression -\pi is not canonical & therefore not bound, it's type cannot be calculated as 'Number', for example.

Oh, I see. That's because there is partial canonicalization applied... Well, this wouldn't help with that case, no. I'm not sure what the expectation would be in this case. Canonicalizing the arguments seems like it could be surprising if you only asked for `Power` canonicalization. But then again, not having the canonicalization apply in this case (but apply if the `Negate` is not there) seems confusing as well.

src/compute-engine/boxed-expression/arithmetic-power.ts
samueltlg: I think that makes sense. Obviously, the most important point there is (being the biggest change):
 `boxed expressions are never bound as a side effect, they are only bound when their canonical form is created (i.e. at construction)`
With the new proposed evaluation frames, this makes more sense/gives way to this change, maybe? (Also, I suppose there is not much need, or it would be a niche one, to inquire into a symbol's value outside of a context? And for that, `.bind()`, I suppose, could be explicitly called beforehand).

The third/final point makes sense. The only potential blocking-point I can consider is whether, for functions, (returning undefined) this does not interfere with anything like the 'handler' functions (such as `sgn`, `type`)? I guess not, since many rely on the function being canonical, anyway...?

For reference - likely you have already seen - I did try to make personal sense of the present behaviour for BoxedSymbols here (https://github.com/cortex-js/compute-engine/commit/a25cc37d147ec9b6ff43ac4cad502a8653287a08#diff-20c94cbe480f0ae94ef00ac68a8904b4da3dfe9248d1f0a111315801905ef803R72): maybe helpful to consider just-in-case you decide not to go with this proposal

samueltlg:
Ok, will probably have to see it in action to fully grasp (its significance).

(Not sure any course of action for second point either; doesn't matter...)

I think I'll just pop-up the missing canonicalPower rule `a^b^c -> a^(b*c)` tomorrow morning, and this block of work should be good to go, I think...

src/compute-engine/boxed-expression/arithmetic-power.ts
arnog: Yes, I did see your comment. Thanks for adding it. I think it's accurate (except that `isSame()` does not require binding AFAIK), and it helped me to realize that the current behavior was perhaps too inconsistent and confusing, hence my motivation to simplify it.

samueltlg: Thought that it could do with a little documentation for reference!

Before I go to sleep, have just pondered whether, it would any-more make sense for a symbol to be bound 'in-place', considering both 1) separation of canonicalization/binding, and 2) evaluation contexts.
I.e., it seems that it is not helpful (at best) for a symbol (non-constant) to be bound in the context of its containing (supposing canonical), expr., this for the large part, only being useful in the context of evaluation/similar?
Maybe this pondering is similar background reasoning to that which has led to the idea of evaluation-contexts...? 
Seems that binding, is unnecessary (until 'last minute'), for non-constant symbols? (it is such that, considering proposed changes, the evaluation-context can be referred to for these, to inquire about their value, alone?)

arnog: I agree, on reflection, binding should be tied to canonicalization.

Binding (canonicalization) may be useful, if not necessary, before the evaluation of symbols, though. Consider a function whose canonicalization depends on whether its arguments are positive, or integers. In that case, the canonicalization handler would need to access this info, regardless of the value of the symbol (and they symbol may not have a value at all).

One case to consider is `["Declare", "i", "'integer'"]`. In this case, the symbol `i` should not be bound automatically, instead the canonical handler of `Declare` should keep `i` as non-canonical, and the evaluation handler should use this unbound symbol to create a new entry in the current scope. Even though this `i` argument is not canonical, the function expression itself is canonical.

samueltlg: Both good examples I did not then consider; I guess even signature validation & potential concomitant inference of unknown symbols (& therefore binding to a type), also applies.

samueltlg:
Think this should be good to go (aside from a small conflict), whenever you're ready...

PR #248, opened 2025-08-01
Patterns substitution fixes

samueltlg:
Sorry that this is later than expected... At least, it's a substantial contribution with some good fixes.

Like usual, will do an inline/source-code review on most outstanding (in literal sense) changes.

There are still one or two remaining issues on the subject of pattern matching (notably, often referenced in source-code comments. Therefore, these likely best reviewed after scanning over changes):

- Using permutations on *patterns* (only) does not, with current behaviour, capture, some cases of commutative-operator operand sequences where it would otherwise be expected.
 A simple case is (extracted from inline doc.) is `['Add', 1, 2, 3, 'x', 5]` matched against pattern `['Add', 1, '__a']`. This does not match (neither prior to, or after changes added here), due to the match being made against the canonicalized input (`['Add', 'x', 1, 2, 3, 4, 5]` (I think).
- This one may more-so just be due to outdated (public) documentation:
Specifying symbols as *strings* for match/replace - via `expr.replace()` - as illustrated within the `Patterns and Rules` guide, does not function *as expected* (i.e. as documentation suggests), due to the 'match' part being parsed as a rule-string: that is, in which symbols (single-character) are parsed as wildcards. 
I.e. the example given listed on this page does not function as indicated, but each rule instead matches/replaces _any_ expression:
```typescript
const expr = ce.box(["Add", ["Multiply", "a", "x"], "b"]);
expr.replace([
    { match: "a", replace: 2 },
    { match: "b", replace: 3 }
  ],
  { recursive: true }
)?.print();
```

And some general notes (again best seen after visiting changes):
- Match permutations (i.e. for commutative expr. operators) could benefit from further optimization.
- (Was going to add this as a final commit, & would not take me much time to add):
  'Pattern' validation ideally to be carried out upon calls to `expr.match()`: such that sequences of (adjacent) regular-sequence wildcards (consider `['List', '__a', '__b', ...]`) are considered 'invalid' (a bit too strong a term).
 In reference to changes: generated permutations skip patterns of this structure, on the basis of these being *redundant* (if not, unpredictable).
 There is the question also, of what the return value should be (of `expr.match()`); `null` does not seem quite fitting. Perhaps something like a 'reason', as returned from `ce.ask()` ...?

And a few future suggestions & *feature requests*:

- Personally, would find it useful to have a `matchPermutations` option (i.e. as applicable to commutative FN's), for `expr.match()`. Sometimes, particularly when wildcards are included in a pattern, I would find desirable to only match exactly as is/as requested.

- Also for *expr.match()*, it would have utility to be able to specify expression-scoped / pertinent *conditions* in a similar way as can be expressed via replacement-rule LaTeX match strings (`x_{numeric} + y_{prime}` for instance.).
  Being able to specify LaTeX-style 'match' strings in the same manner of method 'replace', would not go amiss either...

- For method `replace()`:
  - Arguably, `recursive: true`  should be *default*...? (personal preference/use is for this to be so. Semantically, the sense I get/interpret is 'replace all instances (of)'. Mentionably also, without 'recursive: true', a successful result is nothing but a (top-level) expression re-assignment (assuming assignment of call result) at the Javascript/Typescript level).
  - A way/a syntax variant to *match single-character variables as symbols* (instead of wildcards) in LaTeX strings would be desirable? Could not find a way to do this (but this is of course easily expressible with _boxed_-expressions.
    A possible means could be to use latex ID-prefixes (`\operatorname`, `\mathrm`...), but clearly this would be long-winded. Perhaps alternatively a special minor, syntactical rule to specify this (same level of simplicity as the use of prefix `...` to match sequences, for instance).

 Bit heavy on the text there, but for future reference...!

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -1,11 +1,39 @@
-export function permutations<T>(
-  xs: ReadonlyArray<T>
+/**
+ *
+ * <!--
+ * !@consider?
+ * - In terms of BoxedExpressions - optimizations which are always desirable to take place are
+ * possible...
+ *  ^Perhaps then, a wrapper BoxedExpr. utility for specifying these permutations via 'condition'
+ *  would be apt...?
+ *
+ * - ^If wishing to take adv. of this, the 'condition' callback would likely benefit from a second parameter typed as a collection
+ * ('Set' if enforcing unique) with all hitherto (arbitrary representations) of generated
+ * permutations.
+ *  (See commented snippets within function signature below.)
+ * -->
+ *
+ * @export
+ * @template T
+ * @param xs
+ * @param [condition]
+ * @returns
+ */
+export function permutations<T /* , Y extends any = any */>(
+  xs: ReadonlyArray<T>,
+  condition?: (
+    xs: ReadonlyArray<T> /* , generated: Set<Y> | Set<[Y,T]>? */
+  ) => boolean
+  // cacheKey?: (T) => Y
 ): ReadonlyArray<ReadonlyArray<T>> {
   const result: ReadonlyArray<T>[] = [];
 
-  const permute = (arr, m = []) => {
+  const permute = (arr: T[], m: T[] = []) => {
     if (arr.length === 0) {
-      result.push(m);
+      if (!condition || condition(m)) {
+        // Use spread operator to create a shallow copy of m
+        result.push([...m]);
+      }
     } else {
       for (let i = 0; i < arr.length; i++) {
         const curr = arr.slice();
@@ -15,7 +43,8 @@ export function permutations<T>(
     }
   };
 
-  permute(xs);
+  //@fix: (typing)
+  permute(xs as T[]);
 
   return result;
 }
--- a/src/compute-engine/boxed-expression/match.ts
+++ b/src/compute-engine/boxed-expression/match.ts
@@ -6,7 +6,7 @@ import type {
 
 import { permutations } from '../../common/utils';
 
-import { isWildcard, wildcardName } from './boxed-patterns';
+import { isWildcard, wildcardName, wildcardType } from './boxed-patterns';
 import { isOperatorDef } from './utils';
 
 function hasWildcards(expr: string | BoxedExpression): boolean {
@@ -306,7 +306,33 @@ function matchPermutation(
   options: PatternMatchOptions
 ): BoxedSubstitution | null {
   console.assert(expr.operator === pattern.operator);
-  const patterns = permutations<BoxedExpression>(pattern.ops!);
+
+  // Avoid redundant permutations:
+  // This condition ensures various wildcard sub-permutations - namely a sequence followed by
+  // another sequence or universal, are skipped.
+  // ›Consequently, a pattern such as `['Add', 1, 2, '__a', 4, '__b', 7, '_']`, notably with only 3
+  // non-optional wildcards, yields only *2040* permutations, in contrast to *5040* (i.e. 7!).
+  //!@todo:
+  // - Further optimizations certainly possible here...: consider sub-sequences of expressions with
+  // same values (particularly number/symbol literals). At an extreme, consider ['Add', 1, 1, 1,
+  // 1, 1]. Such a check would reduce permutations from 120 to 1.
+  // - ^Another, may be the req. of a matching qty. of operands (with expr.), in the case of no
+  // sequence wildcards in pattern.
+  const cond = (
+    xs: ReadonlyArray<BoxedExpression> /* , generated: Set<string> */
+  ) =>
+    !xs.some(
+      (x, index) =>
+        isWildcard(x) &&
+        wildcardType(x) === 'Sequence' &&
+        xs[index + 1] &&
+        isWildcard(xs[index + 1]) &&
+        (wildcardType(xs[index + 1]) === 'Wildcard' ||
+          wildcardType(xs[index + 1]) === 'Sequence')
+    );
+
+  const patterns = permutations(pattern.ops!, cond);
+
   for (const pat of patterns) {
     const result = matchArguments(expr, pat, substitution, options);
     if (result !== null) return result;
--- a/src/compute-engine/boxed-expression/match.ts
+++ b/src/compute-engine/boxed-expression/match.ts
@@ -339,39 +339,41 @@ function matchArguments(
 
     if (argName !== null) {
       if (argName.startsWith('__')) {
-        // Match 0 or more expressions (__) or 1 or more (___)
-        let j = 0; // Index in subject
+        // Match 1 or more expressions (__) or 0 or more (___)
+        /** Qty. of operands consumed (of those remaining to be matched). */
+        let j = 0;
         if (patterns[i + 1] === undefined) {
           // No more args in the pattern after, go till the end
-          j = ops.length + 1;
+          j = ops.length;
         } else {
-          // Capture till the next matching arg in the pattern
+          // Capture til' the first matching arg. of the next pattern
           let found = false;
+          const nextPat = patterns[i + 1];
           while (!found && j < ops.length) {
-            found =
-              matchOnce(ops[j], patterns[i + 1], result, options) !== null;
-            j += 1;
+            found = matchOnce(ops[j], nextPat, result, options) !== null;
+            if (!found) j += 1;
           }
-          if (!found && argName.startsWith('___')) return null;
+          // The next pattern does not match against any of the remaining ops.. Unless the next
+          // match is optional, then can assume no overall match.
+          if (!found && !wildcardName(nextPat)?.startsWith('___')) return null;
         }
-
-        // Unless we had a optional wildcard (matching 0 or more), we must have
-        // found at least one match
-        if (!argName.startsWith('___') && j <= 1) return null;
 
         // Determine the value to return for the wildcard
         let value: BoxedExpression;
-        if (j <= 1) {
+        if (j < 1) {
+          // A standard/non-optional Sequence Wildcard with no match...
+          if (!argName.startsWith('___') && j < 1) return null;
+          // Otherwise must be an optional-sequence match: this is permitted.
           if (expr.operator === 'Add') value = ce.Zero;
           else if (expr.operator === 'Multiply') value = ce.One;
           else value = ce.Nothing;
-        } else if (j === 2) {
-          // Capturing a single element
-          if (ops.length === 0) return null;
+        } else if (j === 1) {
+          // Capturing a single element/operand
           value = ops.shift()!;
         } else {
+          // >1 operands captured
           const def = ce.lookupDefinition(expr.operator);
-          const args = ops.splice(0, j - 1);
+          const args = ops.splice(0, j);
           if (def && isOperatorDef(def) && def.operator.associative) {
             value = ce.function(expr.operator, args, { canonical: false });
           } else {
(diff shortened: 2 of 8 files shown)


src/compute-engine/global-types.ts
```diff
@@ -1107,8 +1107,18 @@ export interface BoxedExpression {
...
+   * - If `options.canonical` is set, the *entire expr.* is canonicalized to this degree: whether
+   * the replacement occurs at the top-level, or within/recursively.
+   *
+   * - If otherwise, the *direct replacement will be canonical* if either the 'replaced' expression
+   * is canonical, or the given replacement (- is a BoxedExpression and -) is canonical.
+   * Notably also, if this replacement takes place recursively (not at the top-level), then exprs.
+   * containing the replaced expr. will still however have their (previous) canonical-status
+   * *preserved*... unless this expr. was previously non-canonical, and *replacements have resulted
+   * in canonical operands*. In this case, an expr. meeting this criteria will be updated to
+   * canonical status. (Canonicalization is opportunistic here, in other words).
```
samueltlg, line 1121: Needs your verification and amendment if necessary.
After considering the possibilities here, particularly with recursion, thought that this  strategy makes sense.

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -353,7 +354,10 @@ function parseRulePart(
     );
     return expr;
   }
-  return ce.box(rule, { canonical: options?.canonical ?? false });
+  const canonical =
+    options?.canonical ??
+    (rule instanceof _BoxedExpression ? rule.isCanonical : false);
+  return ce.box(rule, { canonical });
```
samueltlg, line 360: Think that this slightly refined check is a bit more correct ?

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -715,12 +728,11 @@ export function applyRule(
 
   if (canonical && match) {
     const awc = getWildcards(match);
-    const originalMatch = match;
-    match = match.canonical;
```
samueltlg, line 717: This (line) appeared to be non-effective, its removal not affecting any tests, and so removed it.

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -764,6 +776,15 @@ export function applyRule(
...
   }
 
+  // Have a (direct) match: in this case, consider the canonical-status of the replacement, too.
+  if (
+    !canonical &&
+    options?.canonical === undefined &&
+    replace instanceof _BoxedExpression &&
+    replace.isCanonical
+  )
+    canonical = true;
```
samueltlg, line 786: (note: this in reference to the added doc. for method `replace`)

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -419,7 +419,7 @@ function parseRule(
 
         // Check for conditions
         const conditions = parseModifierExpression(parser);
-        if (conditions === null) return null;
+        if (conditions === null) return `${prefix}${id}`;
```
samueltlg, line 422: Believe the fix of this typo has resulted in correct parsing of LaTeX with sequence-syntax (`...n + ...`)

src/compute-engine/boxed-expression/match.ts
```diff
@@ -339,39 +339,41 @@ function matchArguments(
 
     if (argName !== null) {
       if (argName.startsWith('__')) {
-        // Match 0 or more expressions (__) or 1 or more (___)
-        let j = 0; // Index in subject
+        // Match 1 or more expressions (__) or 0 or more (___)
+        /** Qty. of operands consumed (of those remaining to be matched). */
+        let j = 0;
```
samueltlg: Changes here overridden by later commit (68f23c3).
Notably, `j` instead referencing 'qty. of operands consumed', is preserved.

src/compute-engine/boxed-expression/match.ts
```diff
@@ -306,7 +306,33 @@ function matchPermutation(
...
+    xs: ReadonlyArray<BoxedExpression> /* , generated: Set<string> */
+  ) =>
+    !xs.some(
+      (x, index) =>
+        isWildcard(x) &&
+        wildcardType(x) === 'Sequence' &&
+        xs[index + 1] &&
+        isWildcard(xs[index + 1]) &&
+        (wildcardType(xs[index + 1]) === 'Wildcard' ||
+          wildcardType(xs[index + 1]) === 'Sequence')
```
samueltlg: In the subsequent commit, final condition is narrowed to `wildcardType(xs[index + 1]) === 'Sequence'` (only): since filtering out a sequence-wildcard -> universal wildcard sequence (e.g. `['Power', '__', '_']`) a bit too strict (and was breaking some tests).

The reduction of permutations from *5040* (to *2040*) is not so heavy due to this (later change), but still clearly significant

src/common/utils.ts
```diff
@@ -1,11 +1,39 @@
...
+ *
+ * @export
+ * @template T
+ * @param xs
+ * @param [condition]
+ * @returns
+ */
+export function permutations<T /* , Y extends any = any */>(
+  xs: ReadonlyArray<T>,
+  condition?: (
```
samueltlg, line 24: Thought this made sense (see call from `matchPermutation` beneath), but feel free to revert if feeling it doesn't belong

src/compute-engine/boxed-expression/match.ts
```diff
@@ -352,82 +403,217 @@ function matchArguments(
...
-        } else if (j === 1) {
-          // Capturing a single element/operand
-          value = ops.shift()!;
-        } else {
-          // >1 operands captured
-          const def = ce.lookupDefinition(expr.operator);
-          const args = ops.splice(0, j);
-          if (def && isOperatorDef(def) && def.operator.associative) {
-            value = ce.function(expr.operator, args, { canonical: false });
+  return matchRemaining(patterns, substitution);
```
samueltlg, line 410: Recursive-calling necessary due to requirement of attempting variations of operand-capture qty. for sequences (see commit message).

src/compute-engine/global-types.ts
```diff
@@ -1122,6 +1122,9 @@ export interface BoxedExpression {
    *
    * :::info[Note]
    * Applicable to canonical and non-canonical expressions.
+   * 
+   * To specify a match for single symbol (non wildcard), it must be boxed (e.g. `{ match:
+   * ce.box('x'), ... }`), but it is suggested to use method *'subs()'* for this.
```
samueltlg, line 1127: This in reference to opening-PR message: this is the current 'workaround'
(not that it matters since `subs()` can be used)

src/compute-engine/boxed-expression/match.ts
```diff
@@ -299,6 +317,29 @@ function matchVariations(
...
+ * <!--
+ * !@fix?:
+ * - Permutations based on operands of pattern [only] do not match in some scenarios where the
+ * pattern should arguably otherwise return a valid match: specifically in cases involving patterns
+ * with seq.-wildcards. For instance:
+ *   -^A basic example is `['Add', 1, 2, 3, 'x', 5]` matched against pattern `['Add', 1, '__a']`.
+ *   Assuming the input is canonicalized, the resultant expr. against which the pattern is matched
+ *   is `['Add', 'x', 1, 2, 3, 5]`: which cannot be matched against any pattern permutation... (this
+ *   should be expected to be a valid match: with expr. expected to be commutative & therefore
+ *   operands validly taking on any arrangement, canonical-form aside).
```
samueltlg, line 334: Also see opening PR message: this a remaining issue for commutative matching.

samueltlg: (A bit off-topic an observation, but also noticed that for some definitions such as `Set`, these not marked as 'commutative' where they could be)

src/compute-engine/boxed-expression/match.ts
```diff
@@ -352,82 +403,217 @@ function matchArguments(
...
+            let j = isOptionalSeq ? 0 : 1;
+
+            // The next pattern should not be another required sequence wildcard
+            // (^@todo?: validate beforehand; should never be permitted?)
+            console.assert(
+              !(
+                isWildcard(nextPattern) &&
+                wildcardType(nextPattern) === 'Sequence'
+              )
+            );
```
samueltlg, line 462: 'Validation' step as mentioned (for call to `expr.match()`) would still be needed to verify this

src/compute-engine/boxed-expression/match.ts
```diff
@@ -352,82 +403,217 @@ function matchArguments(
...
+
+            while (
+              isWildcard(nextAppPattern) &&
+              wildcardType(nextAppPattern) === 'OptionalSequence'
+            ) {
+              if (!patterns[nextAppPatternIndex + 1]) break;
+              // @note: if pattern has been validated prior, the next should never be a '_'
+              // (universal) or '__' (regular sequence) wildcard
+              nextAppPattern = patterns[++nextAppPatternIndex];
+            }
```
samueltlg, line 480: This is a relatively big added change/check, but believe procedure here is sensible, if you want to verify... ?

src/compute-engine/boxed-expression/match.ts
```diff
@@ -352,82 +403,217 @@ function matchArguments(
...
+              // A sequence wildcard has matched up until the *first* operand that the next (valid) pattern matches.
+              // First try matching remaining ops. against remaining patterns with the sequence
+              // only capturing the operands leading up to this...
+              // Otherwise, continue to see if the next pattern matches (even) further ahead: and
+              // allow this sequence wildcard to capture even more operands.
+              // (^Necessary, for instance, if considering a pattern such as '...a + _n + b' matched
+              // against '3 + 4 + x + b'...: the sequence will have initially captured just '3', but
+              // this will result in a final overall no-match. In this case. allow the sequence to
+              // capture '3 + 4' (finally permitting a 'total' match)
+              while (j <= ops.length) {
```
samueltlg, line 522: This the primary added loop permitting correct/fuller matching of sequences (outlined in this' commit message); appears sensible... ?

samueltlg: Should be end of review

samueltlg:
Also, have seen that there are lots of changes with the recent major release, by the way! Have yet to examine them in more detail (out of interest), but looks good 👏. Were the canonicalization/binding changes included also, and was the primary change here the recognition of an evaluation-scope?
And the change in syntax for types (I think at least signatures: inspired by the scala language, or just went by how you feel/fancied a change 🤔 ? Best

samueltlg:
> Thanks for this PR. Will review it in details soon.
> 
> The canonicalization/binding changes should be incorporated, that is only canonical functions are bound (conversely, bound functions are necessarily canonical).
> 
> There are _lot_ of changes in the latest release. Yes, the addition of proper scoping is one of them (variable capture is not yet implemented, though). There is indeed clearer difference between runtime/evaluation scope and lexical scope (the definitions). This allow proper recursion to work now. One tweak on the type was the syntax for variadic arguments, which used `...xs` previously and now use `xs+` and `xs*`. The motivation for the change is that while using types, I actually could never remember if `...xs` was 1 or more or 0 or more, so I'd figured I would switch to a clearer syntax that I would remember :) Is that what you're referring to? Does Scala really has the same syntax? If so, amusing that we arrived at the same place. The main inspiration for this was actually the syntax for RegEx.

Ah, that's good to hear; I wasn't so sure if those changes (canonicalization) were present from the changelog, but have seen at least some evidence in the source-code. Later, I will try to get more to grips with it: [again] out of interest.
From looking at the Scala documentation some time ago (have not used the language personally), I recall seeing a similar syntax, at least a 'postfix `x`, in the context of function signatures. I may be wrong, but will have to investigate...

Yes, apparently some decent fixes and optimizations present with this one: noticed them (bugs) whilst experimenting & pondering over pattern-matching. Should be worthwhile...

PR #270, opened 2026-01-12
fix: convert small numbers to exact rationals in division ops when possible

### LINKED ISSUE #269: Division by small numbers loses precision
## Description

Dividing by small numbers can lose numerical precision:

### Steps to Reproduce

```ts
const ce = new ComputeEngine();
ce.parse('3 / 0.03').evaluate().latex
```

### Actual Result

```
99.999999...
```

### Expected Result

```
100
```

It should try to treat `0.03` as a decimal, and interpret it as `3 / 100`. Then, such a division will work. E.g. node handles this fine:

```
3 / 0.03
// 100
```

### Environment

**Compute Engine version**: Latest

reteps:
Closes #269 . This fix was written by Claude 4.5 Opus. When trying to divide, it checks if the decimal has a negative exponent. A heuristic is used for `e <= 15` to try and convert this to an exact rational, e.g.

```
1 / 0.000000000000001  -->  exact rational 10¹⁵/1
1 / 0.0000000000000001  --> approximate decimal (existing code)
```

A test is included -- on master, `3 / 0.03` does not yield 100.

### DIFF
--- a/src/compute-engine/numeric-value/big-numeric-value.ts
+++ b/src/compute-engine/numeric-value/big-numeric-value.ts
@@ -193,10 +193,33 @@ export class BigNumericValue extends NumericValue {
     return this.clone({ re: this.decimal.neg(), im: -this.im });
   }
 
-  inv(): BigNumericValue {
+  inv(): NumericValue {
     if (this.isOne) return this;
     if (this.isNegativeOne) return this;
-    if (this.im === 0) return this.clone(this.decimal.pow(-1));
+    if (this.im === 0) {
+      // Check if this is a terminating decimal that can be represented as an exact rational
+      // A decimal with exponent e < 0 can be represented as significand / 10^|e|
+      // e.g., 0.03 (which is 3e-2) becomes 3/100, and its inverse is 100/3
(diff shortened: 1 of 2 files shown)


src/compute-engine/numeric-value/big-numeric-value.ts
```diff
@@ -193,10 +193,33 @@ export class BigNumericValue extends NumericValue {
...
+    if (this.im === 0) {
+      // Check if this is a terminating decimal that can be represented as an exact rational
+      // A decimal with exponent e < 0 can be represented as significand / 10^|e|
+      // e.g., 0.03 (which is 3e-2) becomes 3/100, and its inverse is 100/3
+      const d = this.decimal;
+      if (d.e < 0) {
+        const absExp = -d.e;
+        // Get the significand by shifting the decimal point
+        const significandDecimal = d.times(Decimal.pow(10, absExp));
+        if (significandDecimal.isInteger() && absExp <= 15) {
```
reteps: At 15 digits or less, try to do the improvements. This cutoff is arbitrary.

PR #301, opened 2026-04-11
feat/refactor: 'expr.transform()'; fixes & refactors to 'replace()' and rule-application

samueltlg:
(Code commentary incoming)

Sorry that this was delayed again... Upon thinking it was ready to go, encountered a new set of problems/niggles, requiring a fair amount of investigation.
The change here (fixes) are the sum of quite a lot of tedious debugging!

Most of these are fixes - and an update to 'replace()' - but you may wish to *discard* the head/feature change here (`transform()`). Personally - think it is very useful (not something have added to the personal/bug-fix fork up to this point, but is something will certainly use...)

(For reference, the failing 'compilation-performance' tests, were notably also failing before any of the changes made here))

-------------

Update:
- Key commits are:
  - [fix (workaround): ensure intended canonical - non-canonical -  match pattern ... ](https://github.com/cortex-js/compute-engine/pull/301/changes/57ce5732a131fe1c7ad81990dbf4a190619dc395) (highlights a potential unresolved wildcard/canonicalization bug, as outlined in the commit message)
  - [feat: ReplaceOptions.direction](https://github.com/cortex-js/compute-engine/pull/301/changes/6e116c12420744bc291fd44fd2906b47e3dda3dc) - May not be to your taste (is used for 'expr.transform()'). An alternative could be possible.
  - [Refactor (workaround): during simplification, do not simplify *all* numeric operands of 'lazy' operator definitions](https://github.com/cortex-js/compute-engine/pull/301/changes/5d974d261988513ee9da5602af3cfeea74ced0d7)
    - this is a change bearing relevance to a domain (simplification architecture) still requiring resolution (primarily concerning the 'modularity' of simplify-rules). Lengthy outline in the commit message...
  - [refactor (+fix): update ReplaceOptions.canonical to 'form'; fix: when considering replacement expr. sameness, also consider expression 'form'](https://github.com/cortex-js/compute-engine/pull/301/changes/9c9563017fc3cae2467048b8c451ab542692cb1b) - various comments (pending) throughout here. Also, I thought that jointly upgrading this option, along with the fundamental policy change (applicability to *replacements* only) both offers more flexibility, and is useful for the purposes of *transform()*.
  - [Feat: 'Expression.transform()'](https://github.com/cortex-js/compute-engine/pull/301/changes/2646bf89b2430d7065b4df228477cec401c41f21).
    - The primary contribution. (Note that in addition to adding the 'transform()' function/method, this involves shifting around some key types (`SimplifyOptions`), which would benefit from your review.

Queries & points of contention (after scanning through changes):
----------------------
- 'transform()'
  - The naming of this method is at your discretion. I think 'transform' is on-point enough (designation here borrowed from other libraries).
  - `TransformOptions` option 'targets'
    - Having this as an alternative to 'match' may not be to your taste; but presently this offers a solution and matching route not offered by traditional pattern-matching (- referential-identity; predicate-based). An alternative altogether, or a merge with 'match', may be preferred.
    - !There still exists the problem of this not being able to match against 'cached' (common) engine-exprs (One,Zero,Infinity...): although, for the majority of use-cases, it is hard to imagine the requirement for applying transformations to these expressions anyway (perhaps some cases of 'replace').

Identified further bugs, for reference:
- In a 'replace()' context...:
   - During expression _matching_, the non-canonical expr. is used... Sometimes, however, consequently there can be an unfortunate non-match where a match would otherwise take place. Example:
     - `ce.expr(['Exp', 3]).match(['Exp', '_')])` will __not__ match, due to the non-canonical pattern matching against the canonical variant of `['Exp', 3]`, which notably canonicalizes to a 'Power' FN expr. (`['Power', 'ExponentialE', 3]`).
     - `BoxedString` has a missing `replace()` method (such that replacing always to return 'null').
 
- As highlighted in [fix (workaround): ensure intended canonical - non-canonical -  match pattern...](https://github.com/cortex-js/compute-engine/pull/301/changes/57ce5732a131fe1c7ad81990dbf4a190619dc395):
  - Requesting the _canonical_ variant of wildcard-containing expressions causes entries (BoxedValueDefinition) to be added to _global_ scope for the captured wildcard symbol (e.g. `_` or `_c`).
    - If this point were addressed, this workaround should be able to be retracted.

- A couple of extant issues present with the 'structural' Expression form:
  - Requesting the structural form of canonical vs. non-canonical exprs. presently can produce *differing* output.
    - For the case of a Rational number (representation), for inputs as produced by either `ce.parse('2/5', {form: ...})` or `ce.expr(['Divide', 2, 5], {form: ...})`:
      - In case of boxing/parsing with form 'structural' or 'raw' (and then '*expr.structural*'), the representation is 'Divide' (same in both cases)
       - In case of first obtaining the canonical variant, and then requesting the structural, the result is instead a BoxedNumber (type = *rational*, with an ExactNumericValue value).
     - For the case of an expression with an associative/commutative function head:
       - If obtaining a structural form through requesting this from a canonical expression, then the result will be both sorted (the operands) and flattened.
        - If in contrast boxing/parsing directly as 'structural' - or requesting the 'structural' formfrom of a 'raw' expr. - then expression/operands are neither sorted, or expression flattened.
	  - ^Personally here have identified the source of this to be the fact that the 'canonical' variant has access to the OperatorDefinition: whilst in other cases not.

	
Some possible future directions:
- `expr.subs()` - and possibly one or two others (methods) - could also admit a 'form' specification over 'canonical'.
- If sticking with keeping `expr.transform()`:
  - As per above, TransformOptions 'targets' could be merged with 'match' - or some category of alternative - according to preference.
  - A `verbose` switch for this method: with secondary/meta data returned such as stats on replacement quantity, replacement-expression *instances* (or a map of replaced->replacements), transformation-specific data (e.g. 'steps' for simplify). Perhaps overall expression validity
  - An *async* variant commensurate to 'evaluateAsync()', and others.

### DIFF
--- a/src/compute-engine/boxed-expression/canonical.ts
+++ b/src/compute-engine/boxed-expression/canonical.ts
@@ -75,7 +75,7 @@ export function canonicalForm(
   // Partial canonicalization produces a structural expression, not a fully
   // canonical one. This allows subsequent .canonical calls to perform full
   // canonicalization.
-  if (isFunction(expr) && expr.isCanonical) {
+  if (isFunction(expr) && forms.length > 0) {
     expr = expr.engine.function(expr.operator, [...expr.ops!], {
       form: 'structural',
     });
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -831,13 +831,19 @@ export function applyRule(
   let operandsMatched = false;
 
   if (isFunction(expr) && options?.recursive) {
+    const direction = options?.direction ?? 'left-right';
+    let newOps =
+      direction === 'left-right' ? expr.ops : [...expr.ops].reverse();
+
     // Apply the rule to the operands of the expression
-    const newOps = expr.ops.map((op) => {
+    newOps = newOps.map((op) => {
       const subExpr = applyRule(rule, op, {}, options);
       if (!subExpr) return op;
       operandsMatched = true;
       return subExpr.value;
     });
+
+    if (direction === 'right-left') (newOps as Expression[]).reverse();
 
     // At least one operand (directly or recursively) matched: but continue onwards to match against
     // the top-level expr., test against any 'condition', et cetera.
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -902,7 +902,10 @@ export function applyRule(
       ? replace(expr, sub)
       : replace.subs(sub, { canonical });
 
-  if (!result) return null;
+  if (!result)
+    return operandsMatched
+      ? { value: canonical ? expr.canonical : expr, because }
+      : null;
 
   // To aid in debugging, invoke onMatch when the rule matches
   onMatch?.(rule, expr, result);
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -631,10 +631,7 @@ function boxRule(
   if (typeof condition === 'string') {
     // If the condition is a LaTeX string, it should be a predicate
     // (an expression with a Boolean value).
-    const condPattern =
-      ce.parse(condition, {
-        form: options?.canonical ? 'canonical' : 'raw',
-      }) ?? ce.expr('Nothing');
+    const condPattern = ce.parse(condition) ?? ce.expr('Nothing');
 
     // Substitute any unbound vars in the condition to a wildcard,
     // then evaluate the condition
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -664,16 +664,9 @@ function boxRule(
     );
   }
 
-  // Push a clean scope that only inherits from the system scope (index 0),
-  // not from the global scope or user-defined scopes. This prevents user-defined
-  // symbols (like `x` used as a function name in `x(y+z)`) from interfering with
-  // rule parsing. The system scope contains all built-in definitions.
-  const systemScope = ce.contextStack[0]?.lexicalScope;
-  if (systemScope) {
-    ce.pushScope({ parent: systemScope, bindings: new Map() });
-  } else {
-    ce.pushScope();
-  }
+  // Ensure a clean scope (that only inherits from the system scope) before boxing or parsing:
+  // preventing wildcards & user-defined from inheriting definitions in rules.
+  pushSafeScope(ce);
 
   let matchExpr: Expression | undefined;
   let replaceExpr: Expression | RuleReplaceFunction | RuleFunction | undefined;
@@ -750,6 +743,25 @@ function boxRule(
 }
 
 /**
+ * Push a clean scope - safe for the boxing of rules - that only inherits from the system scope
+ * (index 0), not from the global scope or user-defined scopes. This prevents user-defined symbols
+ * (like `x` used as a function name in `x(y+z)`) from interfering with rule parsing. The system
+ * scope contains all built-in definitions.
+ *
+ * This also crucially prevents wildcards from being given definitions where captured & bound.
+ *
+ * @param ce
+ */
+function pushSafeScope(ce: ComputeEngine) {
+  const systemScope = ce.contextStack[0]?.lexicalScope;
+  if (systemScope) {
+    ce.pushScope({ parent: systemScope, bindings: new Map() });
+  } else {
+    ce.pushScope();
+  }
+}
+
+/**
  * Create a boxed rule set from a collection of non-boxed rules
  */
 export function boxRules(
@@ -806,7 +818,9 @@ export function applyRule(
 
   if (canonical && match) {
     const awc = getWildcards(match);
+    pushSafeScope(ce);
     const canonicalMatch = match.canonical;
+    ce.popScope();
     const bwc = getWildcards(canonicalMatch);
     // If the canonical form of the match loses wildcards, this rule cannot match
     // canonical expressions (they would already be simplified). Skip this rule.
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -18,7 +18,6 @@ import type {
 } from '../global-types';
 
 import {
-  asLatexString,
   isInequalityOperator,
   isRelationalOperator,
 } from '../latex-syntax/utils';
@@ -630,22 +629,19 @@ function boxRule(
   // Normalize the condition to a function
   let condFn: undefined | RuleConditionFunction;
   if (typeof condition === 'string') {
-    const latex = asLatexString(condition);
-    if (latex) {
-      // If the condition is a LaTeX string, it should be a predicate
-      // (an expression with a Boolean value).
-      const condPattern =
-        ce.parse(latex, {
-          form: options?.canonical ? 'canonical' : 'raw',
-        }) ?? ce.expr('Nothing');
+    // If the condition is a LaTeX string, it should be a predicate
+    // (an expression with a Boolean value).
+    const condPattern =
+      ce.parse(condition, {
+        form: options?.canonical ? 'canonical' : 'raw',
+      }) ?? ce.expr('Nothing');
 
-      // Substitute any unbound vars in the condition to a wildcard,
-      // then evaluate the condition
-      condFn = (x: BoxedSubstitution, _ce: ComputeEngine): boolean => {
-        const evaluated = condPattern.subs(x).evaluate();
-        return isSymbol(evaluated, 'True');
-      };
-    }
+    // Substitute any unbound vars in the condition to a wildcard,
+    // then evaluate the condition
+    condFn = (x: BoxedSubstitution, _ce: ComputeEngine): boolean => {
+      const evaluated = condPattern.subs(x).evaluate();
+      return isSymbol(evaluated, 'True');
+    };
   } else {
     if (condition !== undefined && typeof condition !== 'function')
       throw new Error(
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -846,13 +846,15 @@ export function applyRule(
       // The current policy for applying a form according to 'options.form' is for this to apply to
       // *replacements only* (this ultimately allowing for finer control of replacement operations).
       // ...However, if all child operands bear the same form, 'eagerly' assume this form for the
-      // present expression (if this present expression also later matches, form may be updated
-      // according to 'options.form'.)
+      // present expression - provided it has not been explicitly requested that the form be 'raw'.
+      // If this present expression subsequently *directly* matches, the final expression may in any
+      // case be updated according to a specified 'options.form'.
       //(@note: check 'canonical' first, because numbers may be jointly marked as structural and
       //canonical).
-      if (newOps.every((x) => x.isCanonical)) form = 'canonical';
-      else if (newOps.every((x) => x.isStructural)) form = 'structural';
-
+      if (options?.form !== 'raw') {
+        if (newOps.every((x) => x.isCanonical)) form = 'canonical';
+        else if (newOps.every((x) => x.isStructural)) form = 'structural';
+      }
       expr = ce.function(expr.operator, newOps, { form });
     }
   }
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -808,7 +808,8 @@ export function applyRule(
   options?: Readonly<Partial<ReplaceOptions>>
 ): RuleStep | null {
   if (!rule) return null;
-  let canonical = options?.canonical ?? (expr.isCanonical || expr.isStructural);
+  //@todo?: consider 'structural' too (separately); maybe by up-typing 'options.canonical' as 'form'?
+  let canonical = options?.canonical ?? expr.isCanonical;
 
   // eslint-disable-next-line prefer-const
   let { match, replace, condition, id, onMatch, onBeforeMatch } = rule;
--- a/src/compute-engine/boxed-expression/rules.ts
+++ b/src/compute-engine/boxed-expression/rules.ts
@@ -15,6 +15,7 @@ import type {
   Expression,
   ReplaceOptions,
   ExpressionInput,
+  FormOption,
 } from '../global-types';
 
 import {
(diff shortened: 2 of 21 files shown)


src/compute-engine/boxed-expression/rules.ts
```diff
@@ -806,7 +818,9 @@ export function applyRule(
 
   if (canonical && match) {
     const awc = getWildcards(match);
+    pushSafeScope(ce);
```
samueltlg, line 821: (See commit message - this change indeed may not be necessary, if the described present behaviour of the wildcard-symbol canonicalization (global-scope) is not as it should be).

src/compute-engine/types-kernel-serialization.ts
```diff
@@ -153,6 +153,23 @@ export type ReplaceOptions = {
...
+   *
+   * - `'left-right'` reflects *post-order* traversal, (left sub-tree first; depth-descending) (LRN).
+   * - `'right-left'` reflects 'reverse' *post-order* (right sub-tree first; depth-descending) (RLN).
+   *
+   * For both cases traversal is always depth-first, and always visits the root/input expr. last .
+   *
+   * **Default** is: `'left-right'` (standard post-order)
+   *
+   */
+  direction: 'left-right' | 'right-left';
```
samueltlg, line 172: At your discretion! (this is anyway achievable in a workaround-way with custom rule-functions and what-not).

(I did also intend to update the 'once' rule such that it would be typed as something like `'one-rule' | 'one-replacement' | false`: where 'one-rule' would assume the old behaviour of 'true', but 'one-replacement' would command an even stricter policy (currently 'true' applies at the rule level: i.e. a rule may apply multiple times; 'one-replacement' would further restrict this to one replacement/within one rule). This would also somewhat add a '*raison d'être*' for this property ('direction'), too.
However, again, both of these _can_, with a bit of trouble, be achieved through RuleFunctions.)

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -630,22 +629,19 @@ function boxRule(
   // Normalize the condition to a function
   let condFn: undefined | RuleConditionFunction;
   if (typeof condition === 'string') {
-    const latex = asLatexString(condition);
```
samueltlg, line 633: Taking back on this did not *appear* to break anything. Concludes that it was one of either an outdated or mistaken check... the 'asLatexString' call appears to presently require that any string [input] be delimited by an '$' or '$$'

src/compute-engine/boxed-expression/canonical.ts
```diff
@@ -75,7 +75,7 @@ export function canonicalForm(
   // Partial canonicalization produces a structural expression, not a fully
   // canonical one. This allows subsequent .canonical calls to perform full
   // canonicalization.
-  if (isFunction(expr) && expr.isCanonical) {
+  if (isFunction(expr) && forms.length > 0) {
```
samueltlg, line 78: Was not a test to this... but changing this appeared to correct some behaviours whilst outlining the tests for *transform*. Throughout the partial-canonicalization process, appears that sometimes the expression is 'intermediarily' not marked as canonical.

src/compute-engine/boxed-expression/simplify.ts
```diff
@@ -407,7 +407,7 @@ function simplifyNonCommutativeFunction(
   let last = result.at(-1)!.value;
   if (last.isSame(expr)) return steps;
```
samueltlg, line 409: Small oversight... but was inadvertently applying the entire default rule-set to operands! (indeed; no test failures result)

src/compute-engine/boxed-expression/simplify.ts
```diff
@@ -62,7 +62,13 @@ function evaluateNumericSubexpressions(expr: Expression): Expression {
...
 
   // If purely numeric (no unknowns), evaluate the whole expression
-  if (expr.unknowns.length === 0 && BASIC_ARITHMETIC.includes(expr.operator)) {
+  if (
+    expr.unknowns.length === 0 &&
+    BASIC_ARITHMETIC.includes(expr.operator) &&
+    expr.ops.every(
+      (op) => BASIC_ARITHMETIC.includes(op.operator) || !isFunction(op)
+    )
+  ) {
```
samueltlg, line 71: Point of contention, and something to consider (see commit message)

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -998,6 +1032,80 @@ export function replace(
...
+   * For both cases, if neither `x` nor `x2` (nor compared sub-expressions if recursive) is
+   * sructural or canonical, then return `false`.
+   *
+   * **Warning**: will throw an error if it is determined, in case of `recursive !== false`, that
+   * `x` and `x2` are not structurally equivalent/have an identical tree/branching structure.
+   * (It is therefore the responsibility of the caller to ensure this beforehand)
+   *
+   *
+   */
+  function varyingForm(
```
samueltlg, line 1060: This (recursive) check turns out to be necessary if wishing to allow replacements, at-depth within an expression, whereby the structural-sameness remains the same, yet the replacement sub-expression possesses a differing form (structural vs. canonical).
(For the purposes of 'transform' (for structural/canonical) by route of 'replace()', is necessary)

src/compute-engine/boxed-expression/type-guards.ts
```diff
@@ -54,6 +54,17 @@ export function isFunction(
   );
 }
 
+export function assertIsFunction(
```
samueltlg, line 57: This one pretty useful in *test* files as a drop-in assertion check, whilst also offering the benefit of subsequent type-narrowing on the input [expression].

(Have only started/really encountered this typescript operator/feature recently myself!)

src/compute-engine/boxed-expression/utils.ts
```diff
@@ -452,3 +458,23 @@ export function placeholderDef(
...
+ * Get nth (1-based) operand of *expr*; or `null` if this does not exist (or expr is not a
+ * function).
+ * Further, if assert is true (**default**: false), will throw if expr. is not a function.
+ * 
+ * <!--
+ * @todo?: move to 'function-utils'?
+ * -->
+ *
+ */
+export function getOp(
```
samueltlg, line 471: (this one also useful in tests)

src/compute-engine/boxed-expression/rules.ts
```diff
@@ -846,13 +846,15 @@ export function applyRule(
...
-      // according to 'options.form'.)
+      // present expression - provided it has not been explicitly requested that the form be 'raw'.
+      // If this present expression subsequently *directly* matches, the final expression may in any
+      // case be updated according to a specified 'options.form'.
       //(@note: check 'canonical' first, because numbers may be jointly marked as structural and
       //canonical).
-      if (newOps.every((x) => x.isCanonical)) form = 'canonical';
-      else if (newOps.every((x) => x.isStructural)) form = 'structural';
-
+      if (options?.form !== 'raw') {
```
samueltlg, line 854: Another change:
This time round, with the revised 'form' option, there is distinction between explicit provision of value 'raw', and *undefined* (as per this location, and elsewhere local)

src/compute-engine/types-definitions.ts
```diff
@@ -517,42 +511,6 @@ export interface BaseDefinition {
   readonly isConstant?: boolean;
 }
 
-/** Options for `Expression.simplify()`
- *
- * @category Boxed Expression
- */
-export type SimplifyOptions = {
```
samueltlg, line 524: Moved to `types-kernel-evaluation.ts` (and made generic) - in order that it can be employed by options for 'transform()'

src/compute-engine/types-definitions.ts
```diff
@@ -27,9 +24,6 @@ import type {
 export interface ComputeEngine {}
 
 type EvaluateOptions = KernelEvaluateOptions;
-type Rule = KernelRule<Expression, ExpressionInput, ComputeEngine>;
-type BoxedRule = KernelBoxedRule<Expression, ComputeEngine>;
-type BoxedRuleSet = KernelBoxedRuleSet<Expression, ComputeEngine>;
```
samueltlg, line 32: (These re-constructions only previously used by SimplifyOptions (now  re-located))

src/compute-engine/types-expression.ts
```diff
@@ -1277,6 +1282,72 @@ export interface Expression {
     options?: Partial<ReplaceOptions>
   ): null | Expression;
 
+  /**
+   *
+   * Process and transform this expression *recursively* by applying one of a set of predefined
```
samueltlg, line 1287: (This doc.) an outline of justification for this method

src/compute-engine/types-kernel-evaluation.ts
```diff
@@ -95,6 +100,51 @@ export type EvaluateOptions = {
   signal: AbortSignal;
 };
 
+/** Options for `Expression.simplify()`
+ *
+ * @category Boxed Expression
+ */
+
+export type SimplifyOptions<
```
samueltlg, line 108: Possibly not the perfect place for this (i.e. 'evaluate') - but seemed most sensible given the options

src/compute-engine/types-kernel-evaluation.ts
```diff
@@ -229,3 +284,113 @@ export type EvalContext<Expr = unknown, Binding = unknown> = {
...
+    | CanonicalTransformOptions
+    | StructuralTransformOptions
+    | EvaluateTransformOptions
+    | SimplifyTransformOptions<Expr, SemiExpr, CE>
+  );
+
+/** Available transformation types for `Expression.transform()`
+ *
+ * @category Boxed Expression */
+export type Transformation =
```
samueltlg, line 308: (Others could be possible)

src/compute-engine/types-kernel-evaluation.ts
```diff
@@ -229,3 +284,113 @@ export type EvalContext<Expr = unknown, Binding = unknown> = {
...
+   * (Mutually exclusive with `'targets'`). */
+  match?: SemiExpr | LatexString | TransformMatchOptions<Expr, SemiExpr>;
+
+  /** Specify *exact* (referential-identity) transformation targets (sub-expressions), or specify a
+   * predicate for matching. Mutually exclusive with 'match' (pattern-based targeting).
+   *
+   * ::Note
+   * The 'extended' matching routes available here are unique to *transform()* and facilitate
+   * convenient and more expressive matching in the context of recursive traversal.*/
+  targets?: Expr | Expr[] | MatchConditionFunction<Expr>;
```
samueltlg, line 339: Possible point of contention: distinct matching route from 'match'

arnog:
Thanks for putting in the substantial work here, @samueltlg — the bug analyses in the commit messages were really helpful in understanding what each change is solving. There's a lot of good material in this PR.

To make it easier to land incrementally, I've cherry-picked four of the lower-risk fixes from this PR into #303 with regression tests. Authorship is preserved on each cherry-picked commit:

  - `applyRule` no longer returns `null` after operand-level rewrites
  - Rule-application optimization (early exit on canonical-variant wildcard loss)
  - Object-rule condition strings work without `$...$` delimiters
  - Object-rule condition strings are canonicalized so they can evaluate

Once #303 lands, a rebase off `main` here will pick those up cleanly.

For the remaining changes, a few things I'd like to address before the rest is ready to merge:

**Cleanup**

  - `test.only` / `describe.only` markers in `transform.test.ts` (about 13 of them) need to be removed so the tests run in CI.
  - Lint is currently failing with real issues, not just style. The `Multiple exports of name 'LatexString'` (and similar for `LatexDictionaryEntry`, `ParseLatexOptions`, `Parser`, `SerializeLatexOptions`) are coming from a recent refactor on `main` that split `global-types.ts` into multiple `types-*.ts` files — your branch needs to pick a single source for those re-exports. There's also a `no-this-alias` violation around line 720, and prettier wants a sweep.

**Splitting**

The PR is ambitious enough that I'd like to review a few pieces independentlybefore they all land together. Could we split it into roughly:

  1. **`expr.transform()` feature** — the new method, its types, and the `transform.test.ts` file. This is self-contained and the most straightforward piece to review on its own.

  2. **`ReplaceOptions.canonical` → `form`** — this is a public API breaking change without a deprecation path. I'd like to discuss the migration story before merging: at minimum we'd want to keep `canonical` as a deprecated alias for one release, and the type change should be called out in CHANGELOG. Worth a small dedicated PR so this gets visibility.

  3. **The `simplify.ts` BASIC_ARITHMETIC operand guard** — this is the change you flagged as a workaround. It causes three previously-passing tests to be converted to `test.todo` (`10!/(3!*7!)`, two `exp(log(x)±y)` cases). Could you open a tracking issue describing the underlying simplification problem? I'd rather understand the root cause and either fix it properly or accept the workaround with eyes open, rather than have it slip in alongside other changes. Happy to help debug if useful.

  4. **The `canonical.ts` `forms.length > 0` change** (commit `12cabb53`) — this one I'd like to look at separately because it changes how partially-canonicalized expressions are wrapped, and I want to confirm the snapshot drift it causes (e.g. `Power(Multiply(j, 4), -1)` → `Divide(1, Multiply(j, 4))`) is intentional and consistent with the rest of the codebase.

The remaining smaller fixes (`pushSafeScope` consolidation, `direction` option, form-cast policy refinement) can probably go with whichever PR they're most adjacent to.

I realize this is more bureaucracy than ideal. The reason for the splitting ask is that this PR touches enough load-bearing code (rule application, simplification, canonical-form policy, the public `replace()` API) that I want each piece to land with focused review. It also makes bisecting much easier if anything regresses.

Thanks again — looking forward to landing this incrementally.

samueltlg:
Thanks Arno - those requirements sound reasonable and sensible... Indeed, the PR had initial intention of introducing `expr.transform()` perhaps only with one or two accompanying changes or refactors; However,  the attempt to do this resulted in encountering a series of additional issues and requirements - mostly centred in '*replace()*' - which simultaneously demanded attention. (The bureaucracy and preference for splitting are reasonable). Later on tonight (UTC+0) I'll make some headway in making those splits as requested (adhering to 'cleanup' points, and will contribute to CHANGELOG too...). 

(Notably, the 'transform()' set of changes will have to be made finally... with its implementation requiring several of these included fixes/refactors).

samueltlg:
And glad that you are content with keeping `expr.transform()` - I think it's pretty cool 😎  (although, not sure if there is anything you wanted to do between the distinct 'targets' / 'matching' routes - if so just let me know & will make sure to implement).
If you'd find it easier to do the splitting yourself then also do inform; otherwise, I'll have a go at doing so today and tomorrow 👍🏻

samueltlg:
Hello @arnog; have opened two PR's (#305, #306), and one issue (#307) which cover all of the split-points - with exception of the `transform()` commits. I'm sorry that it's a few days late.

Once those are merged, will either force-push the remaining *transform* changes onto this branch, or open a new, final PR.

Should have some other fixes (tangential to these) incoming in a few days time 👍🏻

Thanks

samueltlg:
Can be closed on account of contents being split between #305, #306, #307, #313

PR #312, opened 2026-06-26
feat(latex): add sourceOffsets to parser-generated Error nodes

### LINKED ISSUE #311: [FEATURE] Source offsets on LaTeX parser Error nodes
**Is your feature request related to a problem? Please describe.**

The LaTeX parser reports parse errors as MathJSON `Error` nodes, e.g.

```json
{ "fn": ["Error", { "str": "unexpected-command" }, ["LatexString", { "str": "\\foo" }]] }
```

A consumer can tell *that* something is wrong, and sometimes *what* (`\foo`), but not *where* in the input. To render a precise inline diagnostic (e.g. highlight the offending token in a `MathfieldElement` as a student types), the consumer is forced to either style the whole field or heuristically search the serialized LaTeX for the offending substring — which is ambiguous when the token repeats.

**Describe the solution you'd like**

Populate the existing `MathJsonAttributes.sourceOffsets` field on parser-generated `Error` nodes with the character range of the offending input:

```json
{
  "fn": ["Error", { "str": "unexpected-command" }, ["LatexString", { "str": "\\foo" }]],
  "sourceOffsets": [0, 4]
}
```

- Zero-based, **end-exclusive** offsets into the **serialized LaTeX** (`tokensToString`). For input that round-trips through the tokenizer unchanged (no comments, Unicode normalization, or macro expansion — e.g. editor-generated LaTeX) these equal offsets into the original input.
- **Missing-token** errors (empty `\sqrt{}`, empty `\frac{}{}`) use a **zero-width** range at the parser position.
- A small public helper, `Parser.sourceOffsets(startToken, endToken?)`, lets custom dictionary entries attach offsets to errors they raise; `Parser.error()` uses it internally.
- Offsets propagate through boxing so they also survive `ce.parse(latex).toMathJson()`.

It can be done as a **minimal change (~150 lines, the tokenizer is untouched)** by deriving offsets from the existing `tokensToString` prefix length. A reference implementation + live demo are linked below.

**Describe alternatives you've considered**

- **Status quo (no offsets):** style the whole field on any error, or substring-search the serialized LaTeX for the offending token. Coarse and ambiguous (repeated tokens, nested errors).
- **Thread true source spans through the tokenizer:** a per-token source map carried through every tokenizer transform (comments, NFC, bidi removal, Unicode sub/superscript expansion, macro expansion) would give byte-exact offsets for *any* input, including pasted LaTeX. It works but is a large, invasive change (~330 lines in `tokenizer.ts` alone) with subtle edge cases. The proposed `tokensToString`-prefix approach is exact for editor-generated LaTeX and only drifts for lossy round-trips (comments / normalization / multi-codepoint ZWJ emoji), where the range stays valid but isn't byte-aligned to the original. The minimal version seemed the better default; the heavier one could follow if byte-exact ranges for arbitrary input become a requirement.

**Additional context**

- **Live demo:** https://zojize.github.io/compute-engine/
- **Reference implementation:** [`zojize/compute-engine@codex/latex-error-source-offsets`](https://github.com/zojize/compute-engine/tree/codex/latex-error-source-offsets) (passes the existing `latex-syntax` + `incomplete-expressions` suites with snapshots updated; happy to open a PR).
- **Backward compatibility:** parser `Error` expressions take the object form `{ fn: ["Error", …], sourceOffsets }` instead of the bare array `["Error", …]` when a range is available. Both are valid MathJSON; consumers matching `Array.isArray(expr) && expr[0] === "Error"` should also handle `expr.fn?.[0] === "Error"`.
- **Example usage** (the entire integration for inline highlighting):

  ```js
  import { LatexSyntax } from '@cortex-js/compute-engine';
  
  const ls = new LatexSyntax();
  const ERROR_COLOR = '#dc3545';
  
  // 1. Collect sourceOffsets from parser-generated Error nodes.
  const isOffset = (v) =>
    Array.isArray(v) && v.length === 2 &&
    Number.isInteger(v[0]) && Number.isInteger(v[1]) && v[0] >= 0 && v[1] >= v[0];
  
  function collectErrorOffsets(expr) {
    if (Array.isArray(expr)) return expr.slice(1).flatMap(collectErrorOffsets);
    if (!expr || typeof expr !== 'object') return [];
    const fn = expr.fn;
    if (Array.isArray(fn) && fn[0] === 'Error' && isOffset(expr.sourceOffsets)) {
      return [expr.sourceOffsets];
    }
    return fn ? collectErrorOffsets(fn) : [];
  }
  
  // 2. Map a LaTeX character offset to a Mathfield position by measuring prefixes.
  function offsetToField(mf, target) {
    for (let o = 0; o <= mf.lastOffset; o++) {
      if (mf.getValue(0, o, 'latex-unstyled').length >= target) return o;
    }
    return null;
  }
  
  // 3. On every edit: parse, find error ranges, paint them red.
  let applied = [];
  function highlight(mf) {
    for (const r of applied) mf.applyStyle({ color: 'none' }, { range: r, operation: 'set', silenceNotifications: true });
    applied = [];
  
    const json = ls.parse(mf.getValue('latex-unstyled'));
    for (const [start, end] of collectErrorOffsets(json)) {
      const s = offsetToField(mf, start), e = offsetToField(mf, end);
      if (s === null || e === null || e <= s) continue; // unresolved or zero-width
      mf.applyStyle({ color: ERROR_COLOR }, { range: [s, e], operation: 'set', silenceNotifications: true });
      applied.push([s, e]);
    }
  }
  
  mathfield.addEventListener('input', () => highlight(mathfield));
  ```

zojize:
closes #311

AI usage disclosure: majority of implementation, human reviewed

## Summary

The LaTeX parser reports parse errors as MathJSON `Error` nodes that say *what* went wrong but not *where* in the input. This PR populates the existing `MathJsonAttributes.sourceOffsets` field on parser-generated `Error` nodes with the character range of the offending LaTeX, so consumers can map a parse error back to its source span — e.g. to highlight the exact invalid token inline in a `MathfieldElement`.

```jsonc
// LatexSyntax().parse('\\foo')
{
  "fn": ["Error", { "str": "unexpected-command" }, ["LatexString", { "str": "\\foo" }]],
  "sourceOffsets": [0, 4]
}
```

> **Live demo:** https://zojize.github.io/compute-engine/ — type an invalid expression and the offending token is colored; empty `\sqrt`/`\frac` slots get a red placeholder box. (Closes #<issue-number>.)

## What's exposed

- **`sourceOffsets` on parser `Error` nodes** — zero-based, **end-exclusive** character offsets into the serialized LaTeX (`tokensToString`). For input that round-trips through the tokenizer unchanged (no comments, Unicode normalization, or macro expansion — i.e. editor-generated LaTeX), they equal offsets into the original input string.
- **Missing-token errors** (empty `\sqrt{}`, empty `\frac{}{}`) use a **zero-width** range at the parser position.
- **`Parser.sourceOffsets(startToken, endToken?)`** — public helper converting a token range to that character range, so custom dictionary entries can attach offsets to errors they raise. `Parser.error()` uses it internally.
- **Boxed path** — offsets propagate through boxing/serialization, so they survive `ce.parse(latex).toMathJson()` (selectable via the `metadata` option, included with `metadata: 'all'`).

## Implementation

Deliberately minimal — **`tokenizer.ts` is not touched.** Offsets are derived from the existing `tokensToString` via prefix length, so no per-token source map is threaded through the tokenizer.

| Area | Change |
|---|---|
| `latex-syntax/parse.ts` | `Parser.sourceOffsets()` (`= latex(0, token).length`); `error()` returns `{ fn, sourceOffsets }` |
| `latex-syntax/dictionary/definitions-arithmetic.ts` | empty `\sqrt{}` / `\frac{}{}` operands emit a positioned `missing` error instead of the position-less `MISSING` sentinel |
| `latex-syntax/types.ts`, `math-json/types.ts` | `Parser.sourceOffsets` signature + documented `sourceOffsets` semantics |
| boxed-expression (`box`, `serialize`, `boxed-function`, `abstract-boxed-expression`) + `math-json/utils.ts`, `types-expression.ts`, `types-kernel-serialization.ts` | propagate `sourceOffsets` through `ce.parse().toMathJson()` |

~150 lines of source + test-snapshot updates.

## ⚠️ Behavior change (please review)

Parser `Error` expressions now use the **object form** `{ fn: ["Error", …], sourceOffsets }` instead of the bare array `["Error", …]` whenever a range is available. Both are valid MathJSON, but a consumer matching `Array.isArray(expr) && expr[0] === "Error"` must also handle `expr.fn?.[0] === "Error"`. This is the bulk of the test-snapshot churn in this PR. If you'd prefer this gated behind a parse option rather than on by default, happy to adjust.

## Limitation (documented on the field)

Offsets are measured on `tokensToString(tokenize(input))`, so they drift from the original string only when that round-trip is lossy — comments, Unicode normalization, or multi-codepoint ZWJ emoji. There the range stays valid and monotonic but isn't byte-aligned to the original. Editor-generated LaTeX never hits those paths, so it's exact in practice. A heavier implementation could thread true source spans through the tokenizer if byte-exact ranges for arbitrary pasted input are ever needed.

## Testing

- `latex-syntax/*` and `incomplete-expressions` suites pass; snapshots updated for the object-form errors. New assertions in `errors.test.ts` cover unknown command (`\foo` → `[0,4]`), nested error through `ce.parse().toMathJson()`, unexpected delimiter, and zero-width missing operands.
- Full suite: 10,496 passing. The only failures are two pre-existing, environment-dependent flakes unrelated to this change — `arithmetic.test.ts` (last-digit machine-float drift) and `compile-performance.test.ts` (a ~1MB memory-threshold assertion).
- `tsc --noEmit` clean.

## Notes

- `src/api.md` is **not** regenerated here — running `typedoc`/`concat-md` locally reformats the whole file (toolchain/version drift, ~3k lines of noise). The API documentation lives in the source doc-comments included in this PR; happy to regenerate `api.md` if you'd like it in the same PR or prefer to do it at release time.
- I can add a `CHANGELOG.md` entry under `## [Unreleased] → ### New Features` if that's the expected flow — let me know.

### DIFF
--- a/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
+++ b/src/compute-engine/boxed-expression/abstract-boxed-expression.ts
@@ -132,10 +132,13 @@ export abstract class _BoxedExpression implements Expression {
    *  not generated synthetically
    */
   readonly verbatimLatex?: string;
+  readonly sourceOffsets?: [start: number, end: number];
 
   constructor(ce: ComputeEngine, metadata?: Metadata) {
     this.engine = ce;
     if (metadata?.latex !== undefined) this.verbatimLatex = metadata.latex;
+    if (metadata?.sourceOffsets !== undefined)
+      this.sourceOffsets = metadata.sourceOffsets;
   }
 
   /**
@@ -336,7 +339,7 @@ export abstract class _BoxedExpression implements Expression {
         (typeof options.metadata === 'string' && options.metadata === 'all') ||
         options.metadata?.includes('all')
       ) {
-        defaultOptions.metadata = ['latex', 'wikidata'];
+        defaultOptions.metadata = ['latex', 'wikidata', 'sourceOffsets'];
       } else if (Array.isArray(options.metadata)) {
         defaultOptions.metadata = options.metadata;
       }
--- a/src/compute-engine/boxed-expression/box.ts
+++ b/src/compute-engine/boxed-expression/box.ts
@@ -478,6 +478,11 @@ export function box(
       ? {
           latex: (expr as ExpressionObject & { latex?: string }).latex,
           wikidata: (expr as ExpressionObject & { wikidata?: string }).wikidata,
+          sourceOffsets: (
+            expr as ExpressionObject & {
+              sourceOffsets?: [start: number, end: number];
+            }
+          ).sourceOffsets,
         }
       : undefined;
 
--- a/src/compute-engine/boxed-expression/boxed-function.ts
+++ b/src/compute-engine/boxed-expression/boxed-function.ts
@@ -319,13 +319,23 @@ export class BoxedFunction
         this.isValid ? sortOperands(this._operator, ys) : ys,
         {
           form: 'structural',
+          metadata: {
+            latex: this.verbatimLatex,
+            sourceOffsets: this.sourceOffsets,
+          },
         }
       );
     }
     return this.engine.function(
       this._operator,
       this.ops.map((x) => x.structural),
-      { form: 'structural' }
+      {
+        form: 'structural',
+        metadata: {
+          latex: this.verbatimLatex,
+          sourceOffsets: this.sourceOffsets,
+        },
+      }
     );
   }
 
--- a/src/compute-engine/boxed-expression/serialize.ts
+++ b/src/compute-engine/boxed-expression/serialize.ts
@@ -470,16 +470,28 @@ function serializeJsonFunction(
   // Determine if we have some wikidata metadata
   if (!options.metadata.includes('wikidata')) md.wikidata = '';
 
+  const sourceOffsets = md.sourceOffsets;
+
   //  Is shorthand allowed, and no metadata to include
-  if (!md.latex && !md.wikidata && options.shorthands.includes('function'))
+  if (
+    !md.latex &&
+    !md.wikidata &&
+    !sourceOffsets &&
+    options.shorthands.includes('function')
+  )
     return fn;
 
   // No shorthand allowed, or some metadata to include
-  if (md.latex && md.wikidata)
-    return { fn, latex: md.latex, wikidata: md.wikidata } as MathJsonExpression;
-  if (md.latex) return { fn, latex: md.latex } as MathJsonExpression;
-  if (md.wikidata) return { fn, wikidata: md.wikidata } as MathJsonExpression;
-  return { fn } as MathJsonExpression;
+  const result = { fn } as {
+    fn: MathJsonExpression;
+    latex?: string;
+    wikidata?: string;
+    sourceOffsets?: [start: number, end: number];
+  };
+  if (md.latex) result.latex = md.latex;
+  if (md.wikidata) result.wikidata = md.wikidata;
+  if (sourceOffsets) result.sourceOffsets = sourceOffsets;
+  return result as MathJsonExpression;
 }
 
 function serializeJsonString(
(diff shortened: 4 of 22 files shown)


arnog:
Looks good. I'll add the metadata-gating and take it from here. Thanks for the contribution!

PR #314, opened 2026-06-27
fix: update typecheck script to use npx for TypeScript compilation

samueltlg:
Caught me out! (a global 'tsc' inadvertently running)


arnog:
Good catch

PR #315, opened 2026-06-27
Add differential equation solvers

KingArth0r:
## Summary

Adds an initial, modular differential-equation solving slice for the compute engine:

- registers symbolic `DSolve` in the calculus library
- implements symbolic first-order linear scalar ODE solving in `src/compute-engine/symbolic/differential-equations.ts`
- registers numeric `NDSolve` in the calculus library
- implements fixed-step scalar RK4 support in `src/compute-engine/numerics/differential-equations.ts`
- adds focused tests for symbolic solutions, unsupported inert cases, RK4 sample accuracy, and a Gaussian IVP with a non-elementary antiderivative checked against `Erf`

## Scope / Limitations

This is intentionally narrow so the API and result shape can get feedback before broader ODE support is added.

Current symbolic support handles first-order linear equations of the form `y'(x) + p(x)y(x) = q(x)` and leaves unsupported/nonlinear/higher-order equations inert.

Current numeric support handles explicit scalar first-order IVPs with fixed-step RK4 and returns sample data as a `List` of `[x, y]` pairs. It does not yet implement adaptive RK45, systems, interpolation objects, stiff solvers, implicit equations, or higher-order reductions.

## Validation

- `node_modules/.bin/jest.cmd --config ./config/jest.config.cjs test/compute-engine/differential-equations.test.ts --runInBand --reporters default`
- `node_modules/.bin/jest.cmd --config ./config/jest.config.cjs test/compute-engine/calculus.test.ts --runInBand --reporters default --testNamePattern DERIVATION`
- `node_modules/.bin/madge.cmd --circular --extensions ts src/compute-engine`
- `node_modules/.bin/tsc.cmd --target es2022 --module es2022 --moduleResolution bundler --types node --skipLibCheck -d --allowImportingTsExtensions true --emitDeclarationOnly --ignoreConfig --outDir .typecheck ./src/compute-engine.ts`

### DIFF
--- a/src/compute-engine/differential-equation-utils.ts
+++ b/src/compute-engine/differential-equation-utils.ts
@@ -1,0 +1,47 @@
+import type { Expression, IComputeEngine } from './global-types';
+import { isFunction, isSymbol } from './boxed-expression/type-guards';
+
+export function symbolArg(
+  engine: IComputeEngine,
+  arg: Expression | undefined
+): Expression {
+  if (arg === undefined) return engine.error('missing');
+  if (!isSymbol(arg)) return engine.typeError('symbol', arg.type, arg);
+  return arg;
+}
+
+export function isDependentFunction(
+  expr: Expression,
+  dependentName: string,
+  independentName: string
+): boolean {
+  return (
+    isFunction(expr) &&
+    expr.operator === dependentName &&
+    expr.nops === 1 &&
+    isSymbol(expr.op1, independentName)
+  );
+}
+
+export function isDerivativeOfDependent(
+  expr: Expression,
+  dependentName: string,
+  independentName: string
+): boolean {
+  if (isFunction(expr, 'D')) {
+    return (
+      isDependentFunction(expr.op1, dependentName, independentName) &&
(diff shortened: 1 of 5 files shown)


arnog:
Excellent work!

This looks good overall and fits well with the conventions of the Compute Engine.

A few small things:

1. For the integration constant, use a generated constant name (avoiding `freeVariables` in the expression) instead of hardcoding `C`, or alternatively start with `C` if it is not an already defined identifier in scope, and fallback to a constructed constant name if it is.

2. For `NDSolve`, use `canonicalLimits()` to parse the operand, to handle cases like a simple `Tuple/Tripe` (forms which are also supported by the `Integrate/Sum` operators).

3. I'd suggest that `DSolve` should return a `List`, so that it can represent solution families in the future without changing the API shape.

4. You have a few functions (`isDependentFunction()`, `isDerivativeOfDependent()` and the `symbolArg` closure) that should be promoted to shared helpers to avoid duplication.

5. In `NDSolve` you can use `compiled.success` for a quick bypass.

PR #321, opened 2026-07-12
feat: exact rational DMS parsing and Degrees canonicalization

yelliver:
## Summary

- **parseDMS**: numeric DMS no longer collapses to floating-point degrees; it returns a reduced exact rational `(3600·d + 60·m + s) / 3600` wrapped in `Degrees`.
- **Degrees canonical**: exact conversion to radians now applies to any rational degree (e.g. `19/2°`), not only integers.
- **DMS() canonical**: builds degrees via `ce.number([totalSec, 3600])` instead of a float sum, preserving exactness through the boxing pipeline.
- **Tests**: updated raw-parse expectations to use reduced rationals; evaluation tests assert exact LaTeX via `simplify().latex` instead of floating-point approximations.

## Motivation

DMS angles are naturally exact rationals of a degree: `9°30'` is `19/2°`, not `9.5` as a float. The old pipeline aggregated components with floating-point arithmetic at parse time (`deg + min/60 + sec/3600`), which introduced rounding error before canonicalization.

That mattered downstream: `Degrees` only had an exact rational→radian path for **integer** degrees. A parsed angle like `5°37'30"` became a float, missed the exact branch, and expressions such as converting to radians and simplifying could not reduce to clean fractions (e.g. `π/32`).

This change keeps DMS exact end-to-end - parse as reduced rationals, box `DMS()` the same way, and extend `Degrees` canonicalization to all rational arguments - so degree-to-radian conversion stays symbolically correct.

### DIFF
--- a/src/compute-engine/latex-syntax/dictionary/definitions-arithmetic.ts
+++ b/src/compute-engine/latex-syntax/dictionary/definitions-arithmetic.ts
@@ -14,6 +14,7 @@ import {
   MISSING,
   getSequence,
 } from '../../../math-json/utils.js';
+import { reducedRational } from '../../numerics/rationals.js';
 import {
   Serializer,
   Parser,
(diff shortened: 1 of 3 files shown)


arnog:
Great idea! Thank you for the PR.

PR #322, opened 2026-07-12
fix: nested sqrt on literals not flattening to a single root

yelliver:
## Summary

Fix nested sqrt not flattening when the radicand is a numeric literal.

| | Before | After |
|---|---|---|
| Input | $\sqrt{\sqrt{\sqrt{5}}}$ | $\sqrt{\sqrt{\sqrt{5}}}$ |
| `simplify()` | $\sqrt[4]{\sqrt{5}}$ | $\sqrt[8]{5}$ |
| MathJSON | `Root(Sqrt(5), 4)` | `Root(5, 8)` |

Both forms are mathematically equal ($5^{1/8}$), but the before result kept a nested `Sqrt` inside `Root` instead of flattening to a single root.

**Cause:** For literals, simplify folds `Sqrt(5)` into the exact numeric form $(1)\cdot\sqrt{5}$ before the existing `root(sqrt(x), n)` rule runs. That rule only matched `Sqrt` nodes, so it never fired on the folded form.

**Fix:** Extend `root(sqrt(x), n)` in `simplify-power.ts` to also recognize exact numeric square roots $(1)\cdot\sqrt{a}$.

## Test plan

- [x] `sqrt(sqrt(sqrt(5))) = 5^{1/8}` regression test in `simplify.test.ts`
- [x] Existing `Nested Roots` tests still pass

### DIFF
--- a/src/compute-engine/symbolic/simplify-power.ts
+++ b/src/compute-engine/symbolic/simplify-power.ts
@@ -396,9 +396,23 @@ export function simplifyPower(x: Expression): RuleStep | undefined {
     }
 
     // root(sqrt(x), n) -> x^{1/(2n)} (nth root of square root)
+    // Simplify may fold Sqrt(literal) to exact (1)·√a before this rule runs,
+    // so match both the Sqrt node and that numeric square-root form.
+    let innerBase: Expression | null = null;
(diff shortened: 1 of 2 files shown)


arnog:
Looks good. Thanks!

